PR #8, opened 2017-08-09
Github feedback on build script fix

anwzhang:
Fix for Issue 3 & 6
- Build process does not stop when one of steps is failing
- Build process leaves a gradle process running in the background

### DIFF
--- a/buildprebid.sh
+++ b/buildprebid.sh
@@ -11,6 +11,8 @@
 ######################
 # Helper Methods
 ######################
+set -e
+
 function echoX {
 echo -e "PREBID BUILDLOG: $@"
 }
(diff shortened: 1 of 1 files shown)


buildprebid.sh
```diff
@@ -81,19 +85,18 @@ mkdir $TEMPDIR
...
 
 echoX "Assemble builds"
 cd $LIBDIR
 # clean existing build results, exclude test task, and assemble new release build
-(./gradlew -i -x test build > $LOGPATH/build.log 2>&1 || { echoX "Build failed, check log in $LOGPATH/build.log"; echoX "End Script"; exit 1; } ) &
+(./gradlew -i --no-daemon -x test build > $LOGPATH/build.log 2>&1 || die "Build failed, check log in $LOGPATH/build.log" ) &
 PID=$!
 spinner $PID &
 wait $PID
-echoX "Build succeeded"
```
nhedley: why remove this line?

anwzhang: This line is a bit redundant here. If it fails, then this line will never be run, if the script succeeds and continues there're following lines indicate that.

Issue #9, opened 2017-08-10
ServerConnector HttpURLConnection gives TimeOut exception


niranjannewscorp:
I was trying to implement to PreBid android sdk. `ServerConnector`'s HttpURLConnection has 500 ms as it's timeout and it's very little for an HTTP post request. 

I tried to hit same post request through Postman, it took 5-6 seconds to get response. Hence, I get `SocketTimeOutException` a lot of times. Any workaround for this exception?

Also, sometimes I get  following error sometimes in response:
> { "status": "Prebid cache failed: context deadline exceeded"}

anwzhang:
Hi @niranjannewscorp 

I created this PR https://github.com/prebid/prebid-mobile-android/pull/10 so you're now able to adjust the timeout based on the network condition by setting the Settings.REQUEST_TIME_OUR_MILLIS

For the second error, currently prebid-server is having issues with response cache service, and the team is actively working on the issues. I'll keep you updated when the issues are resolved.

niranjannewscorp:
@anwzhang Thank man for quickly looking into it. How do I set `Settings.REQUEST_TIME_OUR_MILLIS`
 
I am using following code to register my AdUnits
> Prebid.init(application, adUnits, PBS_ACCOUNT_ID);

And `Prebid` class internally calls `ServerConnector` AsynTask class.

anwzhang:
Hi @niranjannewscorp 

PR #10 is closed and merged, the change can be found on the master branch. If you want it from gradle, you have to wait for our next release, which is on next Wednesday. If you want to use the feature now, you can build the lib from source.

I've added a setter for you to override the value, call the following method before `Prebid.init` should work:
`Settings.setConnectionTimeOutMillis(2000)`

Our engineers are still working on the cache failed issue at this point.

anwzhang:
Hi @niranjannewscorp 

All problems in this issue are resolved. For the change in SDK please find it in the latest release: https://github.com/prebid/prebid-mobile-android/releases/tag/0.0.2

basithnewscorp:
Thanks, @anwzhang for the fix. And @niranjannewscorp for making it better

PR #10, opened 2017-08-10
Grant ability to override request time out for http connection

anwzhang:
Issue is from https://github.com/prebid/prebid-mobile-android/issues/9

This fix gives developer the ability to override the time out by changing Settings.REQUEST_TIME_OUR_MILLIS

### DIFF
--- a/PrebidMobile/PrebidServerAdapter/src/main/java/org/prebid/mobile/prebidserver/ServerConnector.java
+++ b/PrebidMobile/PrebidServerAdapter/src/main/java/org/prebid/mobile/prebidserver/ServerConnector.java
@@ -58,7 +58,7 @@ class ServerConnector extends AsyncTask<Object, Object, JSONObject> {
                 conn.setRequestProperty(Settings.COOKIE_HEADER, existingCookie);
             } // todo still pass cookie if limit ad tracking?
             conn.setRequestMethod("POST");
-            conn.setConnectTimeout(Settings.REQUEST_TIME_OUR_MILLIS);
+            conn.setConnectTimeout(Settings.REQUEST_TIME_OUT_MILLIS);
 
             // Add post data
             OutputStreamWriter wr = new OutputStreamWriter(conn.getOutputStream(), "UTF-8");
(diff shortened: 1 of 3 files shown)


PrebidMobile/PrebidServerAdapter/src/main/java/org/prebid/mobile/prebidserver/internal/Settings.java
```diff
@@ -15,7 +15,7 @@
     // connection settings
     public static final String REQUEST_URL_NON_SECURE = "http://prebid.adnxs.com/pbs/v1/auction";
     public static final String REQUEST_URL_SECURE = "https://prebid.adnxs.com/pbs/v1/auction";
-    public static final int REQUEST_TIME_OUR_MILLIS = 500;
+    public static int REQUEST_TIME_OUR_MILLIS = 500;
```
nhedley: If this can be set now, does it still make sense to have it as a constant?

I think if we're going to expose setting the time out in the client facing API, we should write a setter method and make sure its documented, rather than just removing the final here.

anwzhang:
@nhedley setter is added, let me know your thoughts on this.

PrebidMobile/PrebidServerAdapter/src/main/java/org/prebid/mobile/prebidserver/internal/Settings.java
```diff
@@ -15,7 +15,7 @@
     // connection settings
     public static final String REQUEST_URL_NON_SECURE = "http://prebid.adnxs.com/pbs/v1/auction";
     public static final String REQUEST_URL_SECURE = "https://prebid.adnxs.com/pbs/v1/auction";
-    public static final int REQUEST_TIME_OUR_MILLIS = 500;
+    public static int REQUEST_TIME_OUT_MILLIS = 500;
```
nhedley: Is there convention for all caps here? I thought it was all caps implies constant, and camelCase is a variable, but may be wrong.

PrebidMobile/PrebidServerAdapter/src/main/java/org/prebid/mobile/prebidserver/ServerConnector.java
```diff
@@ -58,7 +58,7 @@ protected JSONObject doInBackground(Object... voids) {
                 conn.setRequestProperty(Settings.COOKIE_HEADER, existingCookie);
             } // todo still pass cookie if limit ad tracking?
             conn.setRequestMethod("POST");
-            conn.setConnectTimeout(Settings.REQUEST_TIME_OUR_MILLIS);
+            conn.setConnectTimeout(Settings.REQUEST_TIME_OUT_MILLIS);
```
nhedley: Should we use the getter here?

nhedley:
Let's be sure to add client facing docs on Android if this is something that publishers will use often.

anwzhang:
Currently prebid.org only contains code integration section, not the full API usage documentation. I can add it to the backlog that we need full API usage documentation somewhere.

PR #15, opened 2017-09-18
FB Integration

anwzhang:
This PR includes the implementation for FB demand and unit tests.


PrebidMobile/DemandSDKAdaptersCommon/src/main/java/org/prebid/demandsdkadapters/common/CacheService.java
```diff
@@ -0,0 +1,70 @@
...
+
+    public CacheService(CacheListener listener, String cacheID) {
+        this.listener = listener;
+        this.cacheId = cacheID;
+    }
+
+    @Override
+    protected JSONObject doInBackground(Object... objects) {
+        try {
+            StringBuilder sb = new StringBuilder("http://prebid.adnxs.com/pbc/v1/get?uuid=");
```
nhedley: use /cache endpoint here

PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/Prebid.java
```diff
@@ -236,12 +289,55 @@ private static void handleDFPCustomTargetingUpdate(Object adRequestObj, String a
...
+                    // pass bids that require demand sdk only when rendering is enabled
+                    // save the cache id locally on the ad object for later use
+                    Class mediation_adapter = null;
+                    Set<String> keywords = (Set) callMethodOnObject(adRequestObj, "getKeywords");
+                    for (String keyword : keywords) {
+                        if (PREBID_INTERSTITIAL.equals(keyword)) {
+                            mediation_adapter = getClassFromString(PREBID_DFP_CUSTOM_EVENT_INTERSTITIAL);
+                        } else if (PREBID_BANNER.equals(keyword)) {
+                            mediation_adapter = getClassFromString(PREBID_DFP_CUSTOM_EVENT_BANNER);
+                        }
```
nhedley: Is it possible here to figure out whether or not this is a banner or interstitial based on the adRequestObj? or any other info that doesn't require the publisher to add the additional lines

builder.addCustomEventExtrasBundle(PrebidCustomEventBanner.class, new Bundle());
builder.addKeyword("prebid_banner");

anwzhang: I tested that without adding the bundle for PrebidCustomEventBanner.class on the builder, you can't use reflection or other way to have bundle to be passed. Banner and interstitial ad both uses the same ad request object type, so can't figure out banner or interstitial just based on the object.

nhedley: What about based on the adUnitCode? can we pull the adUnit and therefore the type of ad unit?

anwzhang: That's a good idea.
But builder.addCustomEventExtrasBundle(PrebidCustomEventBanner.class, new Bundle()); has to be there.

PrebidMobile/DemandSDKAdaptersForDFP/src/main/java/org/prebid/demandsdkadapters/dfp/PrebidCustomEventBanner.java
```diff
@@ -0,0 +1,111 @@
...
+public class PrebidCustomEventBanner implements CustomEventBanner, AdListener {
+    private BannerController controller;
+    private CustomEventBannerListener listener;
+
+    @Override
+    public void requestBannerAd(Context context, CustomEventBannerListener customEventBannerListener, String s, AdSize adSize, MediationAdRequest mediationAdRequest, Bundle bundle) {
+        LogUtil.i("Requesting custom event banner through DFP.");
+        this.listener = customEventBannerListener;
+        if (bundle != null) {
+            String cacheId = (String) bundle.get(PREBID_CACHE_ID);
```
nhedley: Do we need to make sure cacheId isn't null here? For example, is there a case where the pub forgets to add the bundle and therefore we aren't able to add the cacheId and the line item still gets here?

PrebidMobile/DemoApp/src/main/java/org/prebid/mobile/demoapp/dummyfragments/FBRequest.java
```diff
@@ -0,0 +1,214 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
+import java.io.UnsupportedEncodingException;
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.URL;
+
+public class FBRequest extends AsyncTask<Object, Object, JSONObject> {
```
nhedley: This file is just for testing right? Can it be removed yet?

PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/Prebid.java
```diff
@@ -182,11 +193,52 @@ private static Object callMethodOnObject(Object object, String methodName, Objec
...
+            }
+            if (bidShouldBeLoadedInDemandSdk) {
+                // pass bids that require demand sdk only when rendering is enabled
+                // save the cache id locally on the ad object for later use
+                Class mediation_adapter = getClassFromString(PREBID_MOPUB_CUSTOM_EVENT_BANNER);
+                if (mediation_adapter != null && PrebidDemandSettings.isDemandEnabled(bidder)) {
+                    for (Pair<String, String> p : keywordPairs) {
+                        keywords.append(p.first).append(":").append(p.second).append(",");
+                        if (PREBID_CACHE_ID.equals(p.first) || PREBID_BIDDER.equals(p.first)) {
+                            Map<String, Object> localExtras = (Map) callMethodOnObject(adViewObj, "getLocalExtras");
```
nhedley: Is there a reason the localExtras need to be set in the loop? Maybe if you pull in cache id above then we don't even need the loop here.

PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/Prebid.java
```diff
@@ -236,12 +287,53 @@ private static void handleDFPCustomTargetingUpdate(Object adRequestObj, String a
...
+                    } else if (AdType.INTERSTITIAL.equals(adUnit.getAdType())) {
+                        mediation_adapter = getClassFromString(PREBID_DFP_CUSTOM_EVENT_INTERSTITIAL);
+                    }
+                    if (mediation_adapter != null && PrebidDemandSettings.isDemandEnabled(bidder)) {
+                        for (Pair<String, String> keywordPair : prebidKeywords) {
+                            bundle.putString(keywordPair.first, keywordPair.second);
+                            usedKeywordKeys.add(keywordPair.first);
+                            // set custom event extras for the requested type
+                            if (mediation_adapter != null) {
+                                Bundle customEventExtras = (Bundle) callMethodOnObject(adRequestObj, "getCustomEventExtrasBundle", mediation_adapter);
```
nhedley: Again, if this doesn't need to be done in the loop, better to loop first and get cacheId and bidder and then set it on DFP customEventExtras.

PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/PrebidDemandSettings.java
```diff
@@ -0,0 +1,124 @@
...
+            if (demand.getBidderName().equals(bidderName)) return true;
+        }
+        return false;
+    }
+
+    public static void enableDemand(Demand demand) throws Exception {
+        String errorMessage = "Prebid SDK uses %s that's not compatible with the version you're using.";
+        switch (demand) {
+            case FACEBOOK:
+                try {
```
nhedley: Pull this block into a separate function to clean it up?

PrebidMobile/PrebidMobileCore/src/test/java/org/prebid/mobile/core/PrebidTest.java
```diff
@@ -0,0 +1,276 @@
...
+    public void setUp() {
+        super.setup();
+        Prebid.setTestServer(MockServer.class.getName());
+
+    }
+
+    @Test
+    public void testPrebidAttachForMoPub() throws Exception {
+        // set up bid response
+        ArrayList<BidResponse> bids = new ArrayList<>();
```
nhedley: Is there a benefit to having all these tests in one block instead of separating them out into smaller tests?

PrebidMobile/PrebidServerAdapter/src/main/java/org/prebid/mobile/prebidserver/PrebidServerAdapter.java
```diff
@@ -374,6 +382,31 @@ else if (context != null
                 device.put(Settings.REQUEST_GEO, geo);
             }
 
+            // ip address
```
nhedley: Can remove

PrebidMobile/PrebidServerAdapter/src/main/java/org/prebid/mobile/prebidserver/PrebidServerAdapter.java
```diff
@@ -388,7 +421,10 @@ else if (context != null
             // os
             device.put(Settings.REQUEST_OS, Settings.os);
             device.put(Settings.REQUEST_OS_VERSION, String.valueOf(Build.VERSION.SDK_INT));
-        } catch (JSONException e) {
+        } catch (
+                JSONException e)
```
nhedley: reformat to have `} catch (JSONException e)` on the same line

PrebidMobile/PrebidServerAdapter/src/main/java/org/prebid/mobile/prebidserver/PrebidServerAdapter.java
```diff
@@ -440,6 +476,16 @@ private JSONObject getUserObject() {
             }
         } catch (JSONException e) {
         }
+        try {
```
nhedley: Let's decide on whether to do this only for FB demand enabled or every time if the class exists.

PrebidMobile/DemandSDKAdaptersCommon/src/main/java/org/prebid/demandsdkadapters/common/BannerController.java
```diff
@@ -0,0 +1,164 @@
...
+    private void loadFacebookBanner(Context context, JSONObject cacheResponse, int width, int height, final AdListener listener) {
+        try {
+            String adm = cacheResponse.getString(PREBID_ADM);
+            JSONObject bid = new JSONObject(adm);
+            String placementID = bid.getString(FACEBOOK_PLACEMENT_ID);
+            Class<?> adViewClass = Class.forName(FACEBOOK_ADVIEW_CLASS);
+            Class<?> adSizeClass = Class.forName(FACEBOOK_ADSIZE_CLASS);
+            Object adSize = null;
+            if (width == 320 && height == 50) {
+                adSize = adSizeClass.getDeclaredField(FACEBOOK_ADSIZE_BANNER_320_50).get(null);
```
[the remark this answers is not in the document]
anwzhang: calling on static variable.

PrebidMobile/DemandSDKAdaptersForDFP/src/main/java/org/prebid/demandsdkadapters/dfp/PrebidCustomEventBanner.java
anwzhang: If pub forgets to add the bundle then bundle would be null this won't be called.

PrebidMobile/DemoApp/src/main/java/org/prebid/mobile/demoapp/dummyfragments/FBRequest.java
anwzhang: Will remove this after testing interstitial.

PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/Prebid.java
anwzhang: MoPub returns a copy of the localExtras not the original object, so the copy of data is needed here.
What do you mean by pull in cache id above?

PrebidMobile/PrebidMobileCore/src/test/java/org/prebid/mobile/core/PrebidTest.java
anwzhang: To test that new keywords are overriding old ones.

PR #19, opened 2017-10-17
Fixed the DFP Interstitial ad

randeepsiddhu:
- Fixed the DFP Interstitial Ad
  Primary ad server was not receiving the keywords due to Prebid.detachUsedBid removing them from the request object.
- Added Constants for Interstitial Fragments
  Use constants instead of hardcoded ad units for consistency.


PrebidMobile/DemoApp/src/main/java/org/prebid/mobile/demoapp/dfpdemofragments/DFPInterstitialFragment.java
```diff
@@ -67,6 +67,5 @@ public void onClick(View v) {
     public void loadInterstitial(View view) {
         Prebid.attachBids(request, Constants.INTERSTITIAL_ADUNIT_ID, getContext());
         mPublisherInterstitialAd.loadAd(request);
-        Prebid.detachUsedBid(request);
```
sandraleon: is there any need to call this at a later time? maybe inside the dfp callbacks?

PR #26, opened 2017-11-21
Cache manager for SDK

anwzhang:
This PR inlcudes a cache manager that helps to cache bids into webview localStorage. Haven't fully tested because pbm.js has to be changed to read the cache data.

### DIFF
--- a/PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/CacheManager.java
+++ b/PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/CacheManager.java
@@ -2,12 +2,15 @@ package org.prebid.mobile.core;
 
 import android.content.Context;
 import android.os.Handler;
+import android.os.Looper;
 import android.text.TextUtils;
 import android.webkit.WebView;
 
+import org.apache.commons.lang3.RandomStringUtils;
+import org.apache.commons.lang3.StringEscapeUtils;
+
 import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.UUID;
 
 public class CacheManager {
     // This is the class that manages three cache instances, one for DFP, one for MoPub and one for SDK rendered ad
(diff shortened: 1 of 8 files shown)


PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/CacheManager.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public static CacheManager getCacheManager() {
+        return cache;
+    }
+
+    public String saveCache(String bid, String format) {
+        if (TextUtils.isEmpty(bid)) {
+            return null;
+        }
+        String cacheId = UUID.randomUUID().toString();
```
nhedley: So for the iOS solution, i'm considering setting the `cacheId` on the server still, even if we don't cache the bid (ie just generating a UUID on server instead of here).

That way we don't have to update the keys on the client side and it will work seamlessly for MoPub iOS solution where we still need to cache.

Thoughts??

PrebidMobile/PrebidServerAdapter/src/main/java/org/prebid/mobile/prebidserver/PrebidServerAdapter.java
```diff
@@ -97,16 +98,18 @@ public void onServerResponded(JSONObject response) {
...
                                 if (responseList == null) {
                                     responseList = new ArrayList<BidResponse>();
                                 }
+                                JSONObject targetingKeywords = bid.getJSONObject(Settings.RESPONSE_TARGETING);
+                                String format = targetingKeywords.getString(Settings.RESPONSE_CREATIVE);
+                                String cacheId = CacheManager.getCacheManager().saveCache(bid.toString(), format);
                                 BidResponse newBid = new BidResponse(bidPrice, cacheId);
                                 newBid.setBidderCode(bidder);
-                                JSONObject targetingKeywords = bid.getJSONObject(Settings.RESPONSE_TARGETING);
+                                newBid.addCustomKeyword(Settings.RESPONSE_CACHE_ID, cacheId);
```
nhedley: So currently we also have the `hb_cache_id_xyzbidder` keywords too. I think we need to consider that use case for deal bids. If we set the cacheId on server side still, that would eliminate the need to have to set all the keywords after we save the bid.

PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/CacheManager.java
anwzhang: I was thinking about having timestamp in the cache id while working on the pbm.js since we need to have a mechanism to expire the bids after a while. If generating the cache id on the server then this won't be able to work.
Thoughts?

nhedley: Synced offline. Generating the cache id client side will work.

nhedley:
Code looks good. Can you add unit tests? let's also wait to merge this until pbm.js is released so that the integration tests pass

anwzhang:
@nhedley sure. Will add.

rmloveland:
@anwzhang are there any API changes for Android users we need to add to the [Android integration docs](http://prebid.org/prebid-mobile/code-integration-android.html)?

(Looks like no based on the diff? But wanted to confirm.)

PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/CacheManager.java
```diff
@@ -0,0 +1,153 @@
...
+import org.apache.commons.lang3.RandomStringUtils;
+import org.apache.commons.lang3.StringEscapeUtils;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+
+public class CacheManager {
+    // This is the class that manages three cache instances, one for DFP, one for MoPub and one for SDK rendered ad
+    private WebView dfpWebCache;
+    private WebView mopubWebCache;
```
nhedley: We don't need mopub related web cache code here anymore right? since we're using server side cache.

anwzhang: Yes I'll remove this.

randeepsiddhu:
@anwzhang @rmloveland 
The change(and deprecation) to init method has to be called out in the Android Integration docs.
The response parsing in PrebidServerAdapter should handle the missing hb_creative_loadtype field to avoid parsing exception, as this field is only added to the top bid.

anwzhang:
@randeepsiddhu thanks for the reminding, will add the change notice and the missing field handling.

PR #40, opened 2018-03-21
CacheManager no longer leaks Contexts and refactors

gubatron:
- No more anonymous runnables with implicit outer references to CacheManager and its Context referencing WebView property
- Reuse Handler instance for setupBidCleanUpRunnable (and its inner Runnable), setupWebCache
- Reuse WebSettings reference when initializing the dfpWebCache settings, Dalvik isn't too good at code optimization
- Removed unnecessary 'long now' instance, only used once when cleaning up the cache
- Suggestion as a next refactor: get rid of one liner methods setupWebCache, setupBidCleanUpRunnable, just invoke the handler.post calls on the constructor

### DIFF
--- a/PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/CacheManager.java
+++ b/PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/CacheManager.java
@@ -4,11 +4,13 @@ import android.content.Context;
 import android.os.Handler;
 import android.os.Looper;
 import android.text.TextUtils;
+import android.webkit.WebSettings;
 import android.webkit.WebView;
 
 import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.commons.lang3.StringEscapeUtils;
 
+import java.lang.ref.WeakReference;
 import java.util.ArrayList;
 import java.util.HashMap;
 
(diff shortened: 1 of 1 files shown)


PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/CacheManager.java
```diff
@@ -140,4 +117,81 @@ private String getCacheForSDK(String cacheId) {
         }
         return null;
     }
+
+    private static class BidCleanUpRunnable implements Runnable {
```
gubatron: inner static classes don't have `.outer` references.

PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/CacheManager.java
```diff
@@ -140,4 +117,81 @@ private String getCacheForSDK(String cacheId) {
...
+        private final WeakReference<CacheManager> cacheManagerWeakRef;
+        private final Handler handler;
+
```
gubatron: and only objects with references to `Context` objects are the ones we make `WeakReference`s of so they can be cleared if their original `Context` instances are no longer in use, this happens quite a lot when `Activities` are destroyed (app minimized, phone rotated, user went home, user locked screen)

PR #41, opened 2018-03-26
reduce dependency only to commons-lang3 and use implementation instead of deprecated compile in gradle

aldenml:
This will help with the integration in other projects, when pulling more code can be a potential problem due to the DEX limitations in android.


gubatron:
sweet!

PR #42, opened 2018-03-27
removed commons-lang3 dependency

aldenml:
The context of this change is try to avoid the `commons-lang3` dependency. Only one left is the `StringEscapeUtils.escapeEcmaScript` call.

### DIFF
--- a/Maven/pom.xml
+++ b/Maven/pom.xml
@@ -28,11 +28,4 @@
     <artifactId>prebid-mobile-sdk</artifactId>
     <version>0.2.0</version>
     <packaging>jar</packaging>
-    <dependencies>
-        <dependency>
-            <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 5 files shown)


anwzhang:
How are you going to remove dependency on `org.apache.commons.lang3.StringEscapeUtils`

aldenml:
that one I need to look at in more details, the fact that `saveCache` is public does not help, since otherwise the JSON string representation should be enough.

aldenml:
@anwzhang I think I found a way to amend this PR with a `escapeEcmaScript` replacement. In the mean time, can you tell me if you are OK with the dependency removal idea?

aldenml:
I transformed the logic from commons lang in a single loop with inline logic. Maybe you have some more data to unit test that I can add. Waiting for your comments.

PR #46, opened 2018-04-11
Add GDPR support in the SDK

anwzhang:


ppuviarasu:
The consent needs to be saved locally n sent for every request... The publisher will provide this params only once & not every time he opens the app.. The publisher is he sends new consent the old one needs to be replaced with the new values

anwzhang:
@ppuviarasu will fix.

PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/TargetingParams.java
```diff
@@ -112,6 +115,48 @@ static void fetchLocationUpdates(Context appContext) {
 
     //region Public APIs
 
+    public static void setGDPRConsentStrings(Context context, ArrayList<String> strings) {
```
ppuviarasu: why do you expect arraylist here... this should only be a string...

PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/TargetingParams.java
```diff
@@ -112,6 +115,48 @@ static void fetchLocationUpdates(Context appContext) {
...
+        if (strings != null && context != null) {
+            SharedPreferences pref = PreferenceManager.getDefaultSharedPreferences(context);
+            SharedPreferences.Editor editor = pref.edit();
+            editor.putString("Prebid_GDPR_consent_strings", strings.toString());
+            editor.apply();
+        }
+    }
+
+    public static ArrayList<String> getGDPSConsentStrings(Context context) {
```
ppuviarasu: consent is sent has string object in openRTB request... why are you splitting it agin to array here?

PrebidMobile/PrebidServerAdapter/src/main/java/org/prebid/mobile/prebidserver/PrebidServerAdapter.java
```diff
@@ -514,9 +519,30 @@ private JSONObject getUserObject() {
             if (!TextUtils.isEmpty(finalKeywords)) {
                 user.put("keywords", finalKeywords);
             }
+            JSONObject ext = new JSONObject();
+            JSONArray consentStrings = new JSONArray();
+            for (String s : TargetingParams.getGDPSConsentStrings(context)) {
```
ppuviarasu: this is rework because of the split done in the targeting params...

PrebidMobile/PrebidServerAdapter/src/main/java/org/prebid/mobile/prebidserver/PrebidServerAdapter.java
```diff
@@ -514,9 +519,30 @@ private JSONObject getUserObject() {
...
         } catch (JSONException e) {
         }
         return user;
     }
+
+    private JSONObject getRegsObject(Context context) {
+        JSONObject regs = new JSONObject();
+        try {
+            JSONObject ext = new JSONObject();
+            if (TargetingParams.isUnderGDPR(context)) {
```
ppuviarasu: what happens if the GDPR is not set by the publisher? by default it should be sent as true not false

PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/TargetingParams.java
```diff
@@ -111,6 +113,47 @@ static void fetchLocationUpdates(Context appContext) {
...
+
+    public static void setGDPRConsentStrings(Context context, String string) {
+        if (!TextUtils.isEmpty(string) && context != null) {
+            SharedPreferences pref = PreferenceManager.getDefaultSharedPreferences(context);
+            SharedPreferences.Editor editor = pref.edit();
+            editor.putString(consentStringKey, string);
+            editor.apply();
+        }
+    }
+
```
Kowshickkarthick: GDPS???

Also ConsentStrings vs ConsentString???

PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/TargetingParams.java
```diff
@@ -111,6 +113,47 @@ static void fetchLocationUpdates(Context appContext) {
     //endregion
 
     //region Public APIs
+    static final String consentStringKey = "Prebid_GDPR_consent_strings";
+
```
Kowshickkarthick: Why write to Shared preference?? and fetch from it each time a request is made?? This might add to the latency in making the ad call each time.
Can this be handled as just instance variable or a combination of shared preference and instance variable.(or may be i am completely wrong)

PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/TargetingParams.java
```diff
@@ -111,6 +113,47 @@ static void fetchLocationUpdates(Context appContext) {
...
+            editor.apply();
+        }
+    }
+
+    public static String getGDPSConsentStrings(Context context) {
+        if (context != null) {
+            SharedPreferences pref = PreferenceManager.getDefaultSharedPreferences(context);
+            if (pref.contains(consentStringKey)) {
+                return pref.getString(consentStringKey, "");
+            }
```
Kowshickkarthick: We need to check for IAB default `IABConsent_ConsentString` as a fallback in else part.

PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/TargetingParams.java
```diff
@@ -111,6 +113,47 @@ static void fetchLocationUpdates(Context appContext) {
...
+            editor.apply();
+        }
+    }
+
+    public static Boolean isUnderGDPR(Context context) {
+        if (context != null) {
+            SharedPreferences pref = PreferenceManager.getDefaultSharedPreferences(context);
+            if (pref.contains(GDPRKey)) {
+                return pref.getBoolean(GDPRKey, false);
+            }
```
Kowshickkarthick: We need to check for IAB default `IABConsent_SubjectToGDPR` as a fallback in else part.

PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/TargetingParams.java
```diff
@@ -111,6 +113,53 @@ static void fetchLocationUpdates(Context appContext) {
...
+            SharedPreferences pref = PreferenceManager.getDefaultSharedPreferences(context);
+            if (pref.contains(APN_CONSENT_STRING_KEY)) {
+                return pref.getString(APN_CONSENT_STRING_KEY, "");
+            } else if (pref.contains(IABConsent_ConsentString)) {
+                return pref.getString(IABConsent_ConsentString, "");
+            }
+        }
+        return null;
+    }
+
```
Kowshickkarthick: Since its Prebid now the namespace needs to be PREBID_GDPR_KEY

PrebidMobile/PrebidMobileCore/src/main/java/org/prebid/mobile/core/TargetingParams.java
```diff
@@ -111,6 +113,53 @@ static void fetchLocationUpdates(Context appContext) {
...
+            editor.apply();
+        }
+    }
+
+    public static Boolean isSubjectToGDPR(Context context) {
+        if (context != null) {
+            SharedPreferences pref = PreferenceManager.getDefaultSharedPreferences(context);
+            if (pref.contains(PREBID_GDPR_KEY)) {
+                return pref.getBoolean(PREBID_GDPR_KEY, false);
+            } else if (pref.contains(IABConsent_SubjectToGDPR)) {
```
Kowshickkarthick: The IAB value is not a boolean its a string and its values can be 
1 - Subject to GDPR
0 - Not Subject to GDPR
nil - unknown or not set

anwzhang: Yes if it doesn't contain the key then return null at the end.

Kowshickkarthick: pref.getBoolean(IABConsent_SubjectToGDPR, false);

I was concerned about this since the getBoolean is not gonna work.  The value present inside the shared preference is not a Boolean but a String.

PR #58, opened 2018-07-25
Fix Price Granularity

AntoineJac:
For Android the request is always passed with:
"ext": { "prebid": { "storedrequest": { "id": "stored-request"}}}

It should be able to check if the price granularity is hardcode in the application and pass the above string if only nothing is declared.


yoalex5:
@AntoineJac Thank you for this userful pull request. It seems good.
I have prepared a PR with some refactoring. Could you please review it
https://github.com/AntoineJac/prebid-mobile-android/pull/1

anwzhang:
I'm afraid I have to close this PR for the following reasons:

No confirmed publisher use case
Prefer not giving the opportunity to hardcode wrong values
For the existing hardcode, I opened an issue to track: #62 
You're welcome to open a new PR to just fix that.

PR #65, opened 2018-08-17
Fix missing prebid.targeting

AntoineJac:


AntoineJac:
"request.ext.prebid.targeting" must be present event is the value is void {}. Or Prebid Server won't set the KV on the response at response.seatbid[i].bid[j].ext.prebid.targeting

AntoineJac:
@ppuviarasu  and @anwzhang , please review

anwzhang:
Hi @AntoineJac ,

I approved this, does iOS need same work too?

AntoineJac:
Hi @anwzhang , thanks a lot. 
iOS should be fine as you have implemented this change in:
https://github.com/prebid/prebid-mobile-ios/pull/90/files

Issue #76, opened 2018-11-30
BidResponse will never get added to the list if `hb_creative_loadtype` is missing


huishan:
https://github.com/prebid/prebid-mobile-android/blob/179099bb153ca0e316bef2ec3bcaa7e3f8bcae7a/PrebidMobile/PrebidServerAdapter/src/main/java/org/prebid/mobile/prebidserver/PrebidServerAdapter.java#L142

According to:
https://github.com/prebid/prebid-server/pull/750/commits/dd8cdb8425f749694fdde0d38b27d61f36429718

loadtype is already removed, need to update for mobile as well

anwzhang:
@huishan created a PR to fix this: https://github.com/prebid/prebid-mobile-android/pull/77

PR #77, opened 2018-12-03
remove hb_creative_loadtype

anwzhang:
https://github.com/prebid/prebid-mobile-android/issues/76


PrebidMobile/PrebidServerAdapter/src/main/java/org/prebid/mobile/prebidserver/PrebidServerAdapter.java
```diff
@@ -139,8 +139,7 @@ public void onServerResponded(JSONObject response) {
                                         if (targetingKeywords != null && targetingKeywords.length() > 0) {
                                             BidResponse newBid = null;
                                             if (Prebid.useLocalCache()) {
-                                                String format = targetingKeywords.getString("hb_creative_loadtype");
-                                                String cacheId = CacheManager.getCacheManager().saveCache(bid.toString(), format);
+                                                String cacheId = CacheManager.getCacheManager().saveCache(bid.toString(), "html");
```
ppuviarasu: Should we change the CacheManager to remove this conditional check for format?

We are now only saving on web

anwzhang: We're upgrading to 1.0, let's leave it like this,

PR #84, opened 2019-02-21
More integration tests on Travis

anwzhang:


anwzhang:
Three issues:
- can't see the launch screen of emulator, tests are failing because can't get focus 
- PBS constantly timeout or no bids, because of the environment 
- Total run time exceeds travis time limit
Recommend running integration test locally

PR #95, opened 2019-03-26
setInv function

AntoineJac:
Regarding to issue:
https://github.com/prebid/prebid-mobile-ios/issues/165

This is the android PR for the function.


AntoineJac:
Thanks all, let stall this until https://github.com/prebid/prebid-mobile-ios/issues/165 has been validated and an intent to implement decided. #

AntoineJac:
Will use the new pass data api

PR #102, opened 2019-04-26
Change host reference

yoalex5:
Default host was changed from Host.APPNEXUS to Host.CUSTOM.
If a publisher does not call setPrebidServerHost() Log error will be printed

Screen Shot 2019-04-26 at 11 38 33 AM

The Doc hint looks like
Screen Shot 2019-04-26 at 11 57 22 AM


yoalex5:
@ppuviarasu Could you please review this PR ?

Issue #111, opened 2019-06-18
cannot set timeout


theAtheA:
In the pre 1.0 client version there was a way to set the client timeout. In the current version there is no public api to override/set the timeout. Could you please reintroduce this feature.
The [PrebidMobile](https://github.com/prebid/prebid-mobile-android/blob/22efd996cb2050b5b0a67dd609ecab38e5814400/PrebidMobile/API1.0/src/main/java/org/prebid/mobile/PrebidMobile.java) class already contains a timeout feature, but it's not public.

bszekely1:
Hi @theAtheA, I will bring this topic back up within the Prebid org. This was talked about some time ago and it may warrant further discussion.

AntoineJac:
@theAtheA , indeed the field should be public. I guess a small typo as be made as public is used for the variable which is useless as also declared as final. Here is the PR for the fix: #112 

For information you can use below code as workaround:


        try {
            Field timeoutMillisField = PrebidMobile.class.getDeclaredField("timeoutMillis");
            timeoutMillisField.setAccessible(true);
            timeoutMillisField.set(null, 3000);
        } catch (NoSuchFieldException e) {
            e.printStackTrace();
        } catch (IllegalAccessException e) {
            e.printStackTrace();
        }

theAtheA:
> Hi @theAtheA, I will bring this topic back up within the Prebid org. This was talked about some time ago and it may warrant further discussion.

As a note: My colleage requested the reimplementation for the same feature in the iOS sdk and it was already discussed and implemented there.

bszekely1:
This will be released in 1.2

PR #112, opened 2019-06-19
Fix prebid timeout editable

AntoineJac:
public should be the field timeoutMillis so it could be changed if needed from the publisher code when integrating the SDK:
"PrebidMobile.timeoutMillis = 1000;"

I removed public for the var "TIMEOUT_MILLIS" which is final so public is useless.
Issue   #111

### DIFF
--- a/PrebidMobile/API1.0/src/main/java/org/prebid/mobile/PrebidMobile.java
+++ b/PrebidMobile/API1.0/src/main/java/org/prebid/mobile/PrebidMobile.java
@@ -22,9 +22,9 @@ import java.lang.ref.WeakReference;
 
 public class PrebidMobile {
 
-    public static final int TIMEOUT_MILLIS = 2_000;
+    static final int TIMEOUT_MILLIS = 2_000;
 
-    static int timeoutMillis = TIMEOUT_MILLIS; // by default use 10000 milliseconds as timeout
+    public static int timeoutMillis = TIMEOUT_MILLIS; // by default use 10000 milliseconds as timeout
     static boolean timeoutMillisUpdated = false;
 
     private PrebidMobile() {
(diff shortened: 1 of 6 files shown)


AntoineJac:
@yoalex5 , for review. Thanks,

AntoineJac:
@ppuviarasu , can you please review this PR? Thanks,

PrebidMobile/API1.0/src/main/java/org/prebid/mobile/PrebidMobile.java
```diff
@@ -22,9 +22,9 @@
 
 public class PrebidMobile {
 
-    public static final int TIMEOUT_MILLIS = 2_000;
+    static final int TIMEOUT_MILLIS = 2_000;
```
yoalex5: In this case we can use `private static final`

AntoineJac: I agree, I have just made the change

PrebidMobile/API1.0/src/main/java/org/prebid/mobile/PrebidMobile.java
```diff
@@ -22,9 +22,9 @@
 
 public class PrebidMobile {
 
-    public static final int TIMEOUT_MILLIS = 2_000;
+    private static final int TIMEOUT_MILLIS = 2_000;
 
-    static int timeoutMillis = TIMEOUT_MILLIS; // by default use 10000 milliseconds as timeout
+    public static int timeoutMillis = TIMEOUT_MILLIS; // by default use 10000 milliseconds as timeout
```
avohraa: If the objective here is to providing publisher's code access to the `timeoutMillis`, then why not expose it using the public facing API. 
For reference:
```
    public static int getTimeoutMillis() {
        return timeoutMillis;
    }

    public static void setTimeoutMillis(int timeoutMillis) {
        PrebidMobile.timeoutMillis = timeoutMillis;
    }
```

Also, please correct the default value written in this comment - `// by default use 10000 milliseconds as timeout`

AntoineJac: Hi @avohraa , I have made the change. Thanks,

PR #113, opened 2019-06-19
[Android][AdManager][Banner] An ad is not rendered properly. Part 2

yoalex5:
[GitHub issue](https://github.com/prebid/prebid-mobile-ios/issues/175)

1. Multisize support was added
2. UnitTests were added
3. UI test was added

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: android
+dist: trusty
 jdk:
   - oraclejdk8
 env:
--- a/PrebidMobile/API1.0/src/main/java/org/prebid/mobile/Util.java
+++ b/PrebidMobile/API1.0/src/main/java/org/prebid/mobile/Util.java
@@ -16,10 +16,12 @@
 
 package org.prebid.mobile;
 
+import android.annotation.TargetApi;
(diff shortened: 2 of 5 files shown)


yoalex5:
The utility method makes the same as [iOS pr](https://github.com/prebid/prebid-mobile-ios/pull/189)

Please do not merge to the master branch
Tomorrow I am going to make full tests on a real device with different AdMAnagerSDK versions.

yoalex5:
I have updated the PR. Also I have tested it on the real device with different AdManagerSDK versions.
The Utility function looks like:
```
Util.findPrebidCreativeSize(dfpAdView, new Util.CreativeSizeCompletionHandler() {
    @Override
    public void onSize(final Util.CreativeSize size) {
        if (size != null) {
            dfpAdView.setAdSizes(new AdSize(size.getWidth(), size.getHeight()));
        }
    }
});

```
Please note that `Util.CreativeSize size` will be always `null` if API < 19 (Currently the min API for the project is 16)

yoalex5:
@ppuviarasu, Could you please review this PR ?

Issue #116, opened 2019-06-27
1.1.1 findPrebidCreativeSize function does not work with Compile error


yoalex5:
## Type of issue
bug

## Description
If add PrebidSDK as Maven artifact AndroidStudio does not see `Util.CreativeSizeCompletionHandler` and `Util.CreativeSize`

<img width="612" alt="Screen Shot 2019-06-27 at 1 28 58 PM" src="https://user-images.githubusercontent.com/8141931/60259308-a8490f00-98df-11e9-82e9-713f74ddf8fb.png">

<img width="279" alt="Screen Shot 2019-06-27 at 1 31 21 PM" src="https://user-images.githubusercontent.com/8141931/60259411-df1f2500-98df-11e9-8894-e7585ef8cd23.png">

## Steps to reproduce
1. Create a new Android project
2. add PrebidSDK Maven artifact (build.gradle(module) -> dependencies scope -> `implementation 'org.prebid:prebid-mobile-sdk:1.1.1'`)
3. Try to use utility function `Util.findPrebidCreativeSize()`

### Expected results
Utility classes(`CreativeSizeCompletionHandler`, `CreativeSize`) should be available from a client side

### Actual results
Static nested classes are not open to use

## Platform details
PrebidSDK v1.1.1

## Other information
 I think that the issue is in proguard

bszekely1:
Resolved with 1.1.2 release

yoalex5:
The API

```
                Util.findPrebidCreativeSize(dfpAdView, new Util.CreativeSizeCompletionHandler() {
                    @Override
                    public void onSize(@Nullable Util.CreativeSize size) {
                        if (size != null) {
                            dfpAdView.setAdSizes(new AdSize(size.getWidth(), size.getHeight()));
                        }
                    }
                });
```

PR #117, opened 2019-07-01
1.1.1 findPrebidCreativeSize function does not work with Compile error

yoalex5:
GitHub issue #116

### DIFF
--- a/PrebidMobile/API1.0/proguard-rules.pro
+++ b/PrebidMobile/API1.0/proguard-rules.pro
@@ -20,6 +20,7 @@
 # hide the original source file name.
 #-renamesourcefileattribute SourceFile
 
+-keepattributes InnerClasses
 -keep public class * {
     public *;
 }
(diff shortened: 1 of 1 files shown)


anwzhang:
Hi, @yoalex5 

Do I need to rebuild the product or just merge this? Should we update the documentation?

yoalex5:
Hello, @anwzhang 

>Do I need to rebuild the product or just merge this? 
@bszekely1, what do you think ? 
From my perspective, it would be great to rebuild the Maven release and GitHub release v1.1.1 with this commit. 

>Should we update the documentation?
We do not need to update the doc

bszekely1:
Lets rebuild under the 1.1.1 tag. No need to increment the version. This release is to have a working 1.1.1 version.

ptomasroos:
Great. This most likely solves the case that we can't set Gender on TargetingParams either. Same error, nested enum's are not exposed or "stripped"

Screenshot 2019-07-03 at 17 14 27

yoalex5:
@ptomasroos, thanks that you pointed it out. I have updated proguard rules

ptomasroos:
Thanks @yoalex5 

Your commit should also solve #121 in that case

ptomasroos:
Looking fwd to a merge and release of this!

PR #132, opened 2019-07-30
Dr. Prebid Android App

eros902002:
This PR contains the following changes:

- Dr. Prebid app for Android

The app is a new module inside the PrebidMobile project

The current version of Dr. Prebid for Android only support testing with MoPub so far. Since Android doesn't have the same mechanism as iOS that let's the app intercept HTTP traffic, we had to find workarounds to make all the tests work on the same way as on iOS. With MoPub it was possible by analysing their open source SDK but no luck so far with DFP.


PrebidMobile/drprebid/src/main/assets/about.html
```diff
@@ -0,0 +1,46 @@
...
+
+<html lang="en-US">
+
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Dr. Prebid</title>
+
+  <style>
+    h1   {color: rgb(59, 136, 195); font-size:20x; font-weight: bold; font-family: -apple-system, BlinkMacSystemFont, sans-serif}
```
jeanstemp: for all the in-app html help files, the font-family should be something for android. I haven't tested yet so I don't know what it looks like with the apple font-family defined.

eros902002: I used the same html files from the Dr. Prebid iOS app. So that's probably why they have the Apple CSS. But I tested them and it looked fine.

yoalex5:
@anwzhang, @ppuviarasu, could you please take a look at this PR ?

anwzhang:
@eros902002 I read this comment on stack overflow: https://stackoverflow.com/a/38679210

Have you tried setting up VPN service in the app?

eros902002:
@anwzhang We did try with the VPN Service but it was a whole mess and it doesn't work well with HTTPS traffic. You'd need to ask the user to install a root certificate which is another whole problem to generate and maintain from the app point of view.

anwzhang:
@eros902002 Thanks for clarifying this.

anwzhang:
@eros902002 I ran the sample app, all seems to be good except you can't see the ad server request sent and key-value targeting sent, etc. Is this because we're not able to intercept the network calls?

eros902002:
@anwzhang Hi. This feature has not been implemented in this app. iOS does have it and it is planned for the next version but we don't have it yet.

We do have access to that info since we use the request and response to test other parts of the flow.

yoalex5:
@anwzhang, can you please approve this PR as a first version and we will add necessary features in next releases of Dr. Prebid ?

yoalex5:
@eros902002, could you please check an error which has failed the TravisCI ?

eros902002:
Yes. On it.

eros902002:
@yoalex5 @anwzhang The TravisCI issue has been fixed. Please feel free to merge the PR whenever it's ok for you.

PR #140, opened 2019-08-22
PBM native

anwzhang:


PrebidMobile/API1.0Demo/src/main/java/org/prebid/mobile/app/DemoActivity.java
```diff
@@ -63,7 +67,99 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
         }
+        adUnit.fetchDemand(adView, new OnCompleteListener() {
+            @Override
+            public void onComplete(ResultCode resultCode) {
+                adView.loadAd();
+            }
+        });
+    }
+
+    void createDFPNative() {
```
yoalex5: Why don't we use `NativeAdUnit` for DFP ?

PrebidMobile/API1.0/src/main/java/org/prebid/mobile/NativeAdUnit.java
```diff
@@ -0,0 +1,197 @@
...
+    static String REQUEST = "request";
+
+
+    public NativeAdUnit(@NonNull String configId) {
+        super(configId, AdType.NATIVE);
+    }
+
+    HashMap<String, Object> requsetConfig = new HashMap<>();
+
+    public void setContext(Integer contextId) {
```
yoalex5: I propose to use `setContextTypeId` to not mix `Android Context` and `NativeAdUnit Context`

PrebidMobile/API1.0/src/main/java/org/prebid/mobile/NativeAdUnit.java
```diff
@@ -0,0 +1,197 @@
...
+import org.json.JSONObject;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+
+/**
+ * For details of the configuration of native imps, please check this documentation:
+ * https://www.iab.com/wp-content/uploads/2018/03/OpenRTB-Native-Ads-Specification-Final-1.2.pdf
+ */
+public class NativeAdUnit extends AdUnit {
```
yoalex5: 1. Magic numbers
I propose to create a documented, predefined named constant variables for a public API that will describe a destination of an argument

For example the `addEventTracker` method has `Integer event` and it is not clear what exactly it means as a result it is simple to make a mistake. 

More over the implementation can be 
``` Java
public static final class EventType {

    static final int IMPRESSION_ID = 1;
    static final int VIEWABLE_MRC50_ID = 2;
    //...

    /**
    * Impression
    */
    public static final EventType IMPRESSION = new EventType(IMPRESSION_ID);

    /**
    * Visible impression using	MRC	definition
    * at 50% in view for 1	second
    */
    public static final EventType VIEWABLE_MRC50 = new EventType(VIEWABLE_MRC50_ID);

    private final int id;

    public EventType(int id) {
        this.id = id;
    }

    public int getId() {
        return id;
    }
}

```

and usage

``` Java
adUnit.addEventTracker(EventType.IMPRESSION, ...);
adUnit.addEventTracker(new NativeAdUnit.EventType(500), ...);

```

PrebidMobile/API1.0/src/main/java/org/prebid/mobile/NativeAdUnit.java
```diff
@@ -0,0 +1,197 @@
...
+        params.put(HEIGHT, h);
+        params.put(WIDTH, w);
+        params.put(REQUIRED, required);
+        if (type == 1 || type == 3) {
+            params.put(TYPE, type);
+        } else {
+            throw new Exception("Unsupported type " + type + " for image");
+        }
+        if (mimes != null) {
+            String mimesString = "";
```
yoalex5: Maybe we can use
``` Java
String mimesString = TextUtils.join(",", mimes);
```

PrebidMobile/API1.0/src/main/java/org/prebid/mobile/NativeAdUnit.java
```diff
@@ -0,0 +1,197 @@
...
+    static String DATA = "data";
+    static String NATIVE = "native";
+    static String REQUEST = "request";
+
+
+    public NativeAdUnit(@NonNull String configId) {
+        super(configId, AdType.NATIVE);
+    }
+
+    HashMap<String, Object> requsetConfig = new HashMap<>();
```
yoalex5: typo fix `requestConfig `

PrebidMobile/API1.0Demo/src/main/java/org/prebid/mobile/app/DemoActivity.java
```diff
@@ -63,7 +67,99 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
         } else if ("MoPub".equals(intent.getStringExtra(Constants.AD_SERVER_NAME)) && "Interstitial".equals(intent.getStringExtra(Constants.AD_TYPE_NAME))) {
             createMoPubInterstitial();
+        } else if ("DFP".equals(intent.getStringExtra(Constants.AD_SERVER_NAME)) && "Native".equals(intent.getStringExtra(Constants.AD_TYPE_NAME))) {
+            createDFPNative();
+        } else if ("MoPub".equals(intent.getStringExtra(Constants.AD_SERVER_NAME)) && "Native".equals(intent.getStringExtra(Constants.AD_TYPE_NAME))) {
+            createMoPubNative();
+        }
+    }
+
+    void createMoPubNative() {
```
yoalex5: I see an PBS error `Invalid request format: No stored imp found for id: 25e17008-5081-4676-94d5-923ced4359d3`

PrebidMobile/API1.0/src/main/java/org/prebid/mobile/NativeAdUnit.java
```diff
@@ -0,0 +1,197 @@
...
+        eventTracker.put(METHODS, methodsJSONArray);
+        if (extObject instanceof JSONArray || extObject instanceof JSONObject) {
+            eventTracker.put(EXT, extObject);
+        }
+        eventTrackers.put(eventTracker);
+        requsetConfig.put(EVENT_TRACKERS, eventTrackers);
+    }
+
+    public void addTitle(Integer len, Boolean required, Object assetExt, Object titleExt) {
+        HashMap<NATIVE_REQUEST_ASSET, HashMap<String, Object>> assets = (HashMap<NATIVE_REQUEST_ASSET, HashMap<String, Object>>) requsetConfig.get("assets");
```
yoalex5: requsetConfig.get(ASSETS)

PrebidMobile/API1.0/src/main/java/org/prebid/mobile/PrebidServerAdapter.java
```diff
@@ -516,14 +516,107 @@ private JSONArray getImp() throws NoContextException {
...
-                } else {
+                } else if (requestParams.getAdType().equals(AdType.BANNER)) {
                     JSONObject banner = new JSONObject();
                     JSONArray format = new JSONArray();
                     for (AdSize size : requestParams.getAdSizes()) {
                         format.put(new JSONObject().put("w", size.getWidth()).put("h", size.getHeight()));
                     }
                     banner.put("format", format);
                     imp.put("banner", banner);
+                } else {
```
yoalex5: `else if (requestParams.getAdType().equals(AdType.NATIVE)) {`

PrebidMobile/API1.0/src/main/java/org/prebid/mobile/NativeAdUnit.java
yoalex5: It is can not be an enum because usually a type can contain a user defined value. For example `7.6 Event Types Table` has `500+ exchange-specific` as a result a value can be 500, 501, 502...

I think that we should create separate classes or just a predefined named constants for every type(e.g. `EventType`, `Image Type`, `Data Type`, `mimes`,  `methods`...)

PrebidMobile/API1.0Demo/src/main/java/org/prebid/mobile/app/DemoActivity.java
yoalex5: You are right. I used Rubicon host. Currently I see successful response from PBS but MoPub response is empty

``` JSON
{
"ad-responses": [{
    "content": "",
    "metadata": {
        "x-adtype": "clear",
        "x-backfill": "clear",
        "x-refreshtime": 999999
        }
    }],
    "x-request-id": "edc0d651ef8747e2bcfdad8c37f327cb_0090428a006172cd"
}

```

PrebidMobile/API1.0/src/main/java/org/prebid/mobile/NativeAdUnit.java
```diff
@@ -0,0 +1,371 @@
...
+                this.id = id;
+            }
+        }
+    }
+
+    public void setContextSubType(CONTEXTSUBTYPE type) {
+        requestConfig.put(CONTEXT_SUB_TYPE, type.getID());
+    }
+
+    public enum PLACEMENTTYPE {
```
yoalex5: I propose to use `enum PLACEMENT_TYPE` as name of enum and `String PLACEMENT_TYPE_NAME`. The same for `CONTEXTSUBTYPE`.

PrebidMobile/API1.0Demo/src/main/java/org/prebid/mobile/app/DemoActivity.java
```diff
@@ -66,7 +70,113 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
             createMoPubBanner(intent.getStringExtra(Constants.AD_SIZE_NAME));
         } else if ("MoPub".equals(intent.getStringExtra(Constants.AD_SERVER_NAME)) && "Interstitial".equals(intent.getStringExtra(Constants.AD_TYPE_NAME))) {
             createMoPubInterstitial();
+        } else if ("DFP".equals(intent.getStringExtra(Constants.AD_SERVER_NAME)) && "Native".equals(intent.getStringExtra(Constants.AD_TYPE_NAME))) {
```
yoalex5: I still can not get successful result neither from DFP nor MoPub. 
PBS - `<head><title>400 Bad Request</title></head>`

anwzhang: Some configuration was not right and something's not working with PBS, trying to find out why now.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidServerAdapter.java
```diff
@@ -560,14 +559,117 @@ private JSONArray getImp() throws NoContextException {
...
+                    JSONArray trackers = (JSONArray) requestConfig.get(NativeAdUnit.EVENT_TRACKERS);
+                    if (trackers != null && trackers.length() > 0) {
+                        request.put(NativeAdUnit.EVENT_TRACKERS, requestConfig.get(NativeAdUnit.EVENT_TRACKERS));
+                    }
+                    if (requestConfig.get(NativeAdUnit.PRIVACY) != null) {
+                        request.put(NativeAdUnit.PRIVACY, requestConfig.get(NativeAdUnit.PRIVACY));
+                    }
+                    if (requestConfig.get(NativeAdUnit.EXT) != null) {
+                        request.put(NativeAdUnit.EXT, requestConfig.get(NativeAdUnit.EXT));
+                    }
```
yoalex5: It is safe to remove `if` (null check) before putting something into JSONObject.

anwzhang: This is semi-final of the PR, @ppuviarasu is working on iOS version, I'll wait on feedbacks from her implementation.
I'll update based on your feedback too.

yoalex5:
Please review #157

Example/PrebidDemoJava/src/main/AndroidManifest.xml
```diff
@@ -11,14 +11,14 @@
     <!-- Optional permissions. Used for MRAID 2.0 storePicture ads -->
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
     <application
-        android:networkSecurityConfig="@xml/network_security_config"
```
yoalex5: What is the reason of removing `android:networkSecurityConfig` ?

Example/PrebidDemoJava/src/main/java/org/prebid/mobile/app/CustomApplication.java
```diff
@@ -39,18 +40,19 @@
...
-                .withNetworksToInit(networksToInit)
                 .build();
         MoPub.initializeSdk(this, sdkConfiguration, null);
         //set Prebid Mobile global Settings
         //region PrebidMobile API
-        PrebidMobile.setPrebidServerAccountId(Constants.PBS_ACCOUNT_ID);
-        PrebidMobile.setPrebidServerHost(Host.APPNEXUS);
+        PrebidMobile.setPrebidServerAccountId("account");
+        Host custom = Host.CUSTOM;
+        custom.setHostUrl("http://10.0.2.2:8000/openrtb2/auction");
```
yoalex5: Is it necessary to use a custom address ?

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/NativeAsset.java
```diff
@@ -0,0 +1,24 @@
+package org.prebid.mobile;
+
+abstract class NativeAsset {
+    enum REQUEST_ASSET {
+        TITLE,
+        IMAGE,
+        DATA
+    }
```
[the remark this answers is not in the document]
anwzhang: No we're not supporting native video, I'll check with @ppuviarasu see what she thinks.

Example/PrebidDemoJava/src/main/AndroidManifest.xml
anwzhang: Mistake, will add it back.

Example/PrebidDemoJava/src/main/java/org/prebid/mobile/app/CustomApplication.java
anwzhang: This is for testing. Will change it back after end to end testing

Example/PrebidDemoJava/src/main/AndroidManifest.xml
anwzhang: Switched to simple config: android:usesCleartextTraffic="true"

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidServerAdapter.java
anwzhang: If it's null, the generated JSON object will just have a empty field right?

anwzhang:
Hi @yoalex5, I fixed the merge conflicts, still working on aligning the APIs. Can you take a look at the demo app? I don't see Video showing there.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidServerAdapter.java
yoalex5: doc says:
>If the value is null, any existing mapping for name is removed.

yoalex5:
> Hi @yoalex5, I fixed the merge conflicts, still working on aligning the APIs. Can you take a look at the demo app? I don't see Video showing there.

Hello @anwzhang, I checked demo app using your branch and `Banner Video` and `Interstitial Video` work as expected. 

Could you please share a Charles log ?

PR #106, opened 2019-05-31
Travis CI does not work

yoalex5:
GitHub issue #105 

1. Travis yml file was rewritten and contains script from [butterknife](https://github.com/JakeWharton/butterknife/blob/master/.travis.yml) repository
2. EMULATOR_API_LEVEL was changed to version 18
3. `testprebid.sh` script was divided to `testPrebidSDK.sh(Unit tests)` and `testPrebidDemo.sh(UI tests)` scripts. It brings flexibility of running scripts and returning exit codes 
4. `testPrebidSDK.sh(Unit tests)` show console output (display test results)
5. `testPrebidDemo.sh(UI tests)`  runs all instrumentation tests and they can be failed.
6. Run all UI tests instead of running only one SanityTest test

Note: Current pipeline implies that the final CI status depends only on  `testPrebidSDK.sh(Unit tests)`. Current implementation of UI tests do not allow us to include them to strict mode because sometimes they failed. When UI tests are fixed we will update the pipeline. The same logic we stick in[ iOS version](https://github.com/prebid/prebid-mobile-ios/issues/195)

PR #144, opened 2019-09-06
release 1.2 version update

anwzhang:


PrebidMobile/API1.0Demo/build.gradle
```diff
@@ -31,15 +31,15 @@ android {
     }
 }
 
-project.gradle.taskGraph.whenReady {
-    connectedSourceCodeDebugAndroidTest {
-        ignoreFailures = true
-    }
-}
+//project.gradle.taskGraph.whenReady {
```
yoalex5: I have added this setting as a part of #106 

The main reason that DemoApp includes UI tests that rely on Internet connection, real PBS responses etc. As a result we can not guarantee that all tests always are passed. 

We should overview this type of tests.
Please uncomment these lines for now. It should solve TravisCI fail

Issue #101, opened 2019-04-25
How to initialize post data before the first request?


kdravolin:
This is method of getting post data in PrebidServerAdapter:
```
        private JSONObject getPostData() throws NoContextException {
            Context context = PrebidMobile.getApplicationContext();
            if (context != null) {
                AdvertisingIDUtil.retrieveAndSetAAID(context);
                PrebidServerSettings.update(context);
            }
            ...
```
According the code, `AdvertisingIDUtil.retrieveAndSetAAID` and `PrebidServerSettings.update` methods are called asynchronously during a request. So the first request doesn't contain some data e.g. "ifa" and "ua".

Is there any way to initialize this params before the first request?

derN3rd:
Hey, we are having the same issues and seeing that most of our prebid-mobile requests are missing ifa and ua (therefore we don't get any bids).

Are they any news/workarounds for this?

Thanks in advance

anwzhang:
@derN3rd are you seeing missing ifa and ua only in the first calls or most calls?

derN3rd:
Sorry for the late response.

We only see this missing in the first calls.

anwzhang:
Hi @derN3rd @yoalex5 ,

I think there're two solutions:
1. consolidate the following methods into one `initSDK()` method (with parameters)
```
PrebidMobile.setPrebidServerHost(Host.APPNEXUS);
PrebidMobile.setPrebidServerAccountId(Constants.PBS_ACCOUNT_ID);
PrebidMobile.setShareGeoLocation(true);
PrebidMobile.setApplicationContext(getApplicationContext());
```
Then do the initial setup of aaid inside the `initSDK()`, thus we can provide a callback to the user that SDK is ready, but this requires API breaking changes.
2. When a user uses PrebidMobile.setApplicationContext(getApplicationContext()), we immediately starts fetching aaid, etc. This is only internal change.

Thoughts?

yoalex5:
Hello @anwzhang,

I like the second variant because it is easy and does not require API changes

PR #148, opened 2019-10-02
Expose setPrebidUserAaid, getPrebidUserAaid and setPrebidUserAgent

derN3rd:
This PR exposes functions to set the idfa and user agent, as a workaround for https://github.com/prebid/prebid-mobile-android/issues/101

Tested this in our react-native app and the first prebid request therefore has the data, we set before

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdvertisingIDUtil.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdvertisingIDUtil.java
@@ -38,11 +38,11 @@ class AdvertisingIDUtil {
         return limitAdTracking;
     }
 
-    private static synchronized void setAAID(String aaid) {
+    static synchronized void setAAID(String aaid) {
         AdvertisingIDUtil.aaid = aaid;
     }
 
-    private static synchronized void setLimitAdTracking(boolean limitAdTracking) {
+    static synchronized void setLimitAdTracking(boolean limitAdTracking) {
         AdvertisingIDUtil.limitAdTracking = limitAdTracking;
     }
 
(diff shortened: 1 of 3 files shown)


yoalex5:
Hello @derN3rd , thanks for your PR. 

As I see you propose just to expose SDK API to define `aaid`, `limitAdTracking`, `userAgent ` for a publisher. I suppose that they were hidden on purpose. 

I think we should come up with an approach to setup these field internally in the SDK.

@ppuviarasu, what do you think ?

derN3rd:
I agree this is only a workaround (or a way, to set these properties if you already have them).

As mentioned in the issue #101 aaid and userAgent get set async, which might be too late for the first requests, which causes no bids on almost every first request made. But with this exposed functions I set them directly on initialization and so the first request has aaid and userAgent.

As far as I see this is caused by https://github.com/prebid/prebid-mobile-android/blob/644b9cfc77314cfd3204cdeb742285087ed2ee37/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdvertisingIDUtil.java#L83-L96 being async and https://github.com/prebid/prebid-mobile-android/blob/644b9cfc77314cfd3204cdeb742285087ed2ee37/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidServerSettings.java#L80-L91 which also doesn't finish before the first request.

So why not letting the dev set the two values manually if he already has them (in state or something) instead of waiting on their own async tasks and leading to failing ad requests?

anwzhang:
@derN3rd this is resolved by #169 

I'll close this PR once you confirm it.

Jobeso:
We tried out #169 and it works fine so you can close this pr. Thanks for taking care of it.

PR #151, opened 2019-10-22
Video format support (Out-of-the-box)

yoalex5:
GitHub issue #146

API
```Java
new VideoAdUnit("1001_test_video", 300, 250, VideoAdUnit.PlacementType.IN_BANNER);
new VideoInterstitialAdUnit("1001_test_video");
```

`VideoAdUnit` has next types:
```
IN_BANNER
IN_ARTICLE
IN_FEED
``` 

OpennRTB request:
``` XML
{
    "imp": [{
        "video": {
            "w": 640,
            "h": 480,
            "startdelay": 0,
            "protocols": [7],
            "playbackmethod": [2],
            "mimes": ["video/mp4"],
            "placement": 2 <--2, 3, 4 for VideoAdUnit and 5 for VideoInterstitialAdUnit-->
        },
    }],
    "ext": {
        "prebid": {
            "cache": {
                "vastxml": {}
            },
        },
    },
    "instl": 1 <--Only for Interstitial-->
}
```

### DIFF
--- a/Example/PrebidDemoJava/src/main/java/org/prebid/mobile/app/DemoActivity.java
+++ b/Example/PrebidDemoJava/src/main/java/org/prebid/mobile/app/DemoActivity.java
@@ -38,11 +38,14 @@ import com.mopub.mobileads.MoPubView;
 
 import org.prebid.mobile.AdUnit;
 import org.prebid.mobile.BannerAdUnit;
+import org.prebid.mobile.Host;
 import org.prebid.mobile.InterstitialAdUnit;
 import org.prebid.mobile.OnCompleteListener;
 import org.prebid.mobile.PrebidMobile;
 import org.prebid.mobile.ResultCode;
 import org.prebid.mobile.TargetingParams;
+import org.prebid.mobile.VideoAdUnit;
+import org.prebid.mobile.VideoInterstitialAdUnit;
 import org.prebid.mobile.addendum.AdViewUtils;
 import org.prebid.mobile.addendum.PbFindSizeError;
 
@@ -53,6 +56,11 @@ public class DemoActivity extends AppCompatActivity {
     int refreshCount;
     AdUnit adUnit;
     ResultCode resultCode;
+
+    private PublisherAdView amBanner;
+    private PublisherInterstitialAd amInterstitial;
+
+    private MoPubInterstitial mpInterstitial;
 
     @Override
     protected void onCreate(@Nullable Bundle savedInstanceState) {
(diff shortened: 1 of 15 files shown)


PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidServerAdapter.java
```diff
@@ -560,14 +573,38 @@ private JSONArray getImp() throws NoContextException {
...
                     for (AdSize size : requestParams.getAdSizes()) {
                         format.put(new JSONObject().put("w", size.getWidth()).put("h", size.getHeight()));
                     }
                     banner.put("format", format);
                     imp.put("banner", banner);
+                } else if (adType.equals(AdType.VIDEO) || adType.equals(AdType.VIDEO_INTERSTITIAL)) {
+
+                    JSONObject video = new JSONObject();
+                    video.put("mimes", new JSONArray().put("video/mp4"));
+                    video.put("protocols", new JSONArray().put(7));
```
anwzhang: Should this be configurable by the publisher? I also read the doc that there should be more fields.

bszekely1: Typically a publisher sets these to what is supported by the video player. Since we are providing the video player support, we are defining these values for them. If they wish to override them, in the subsequent release, we can read them from Prebid Server using this Intent to implement where a publisher can set them using Stored Requests.:
 https://github.com/prebid/prebid-server/issues/1075

The only real setting I imagine a publisher changing is playbackmethod and possibly protocols. This first pass we are hardcoding the values with the goal of keeping all these attributes in Stored Request for more flexibility vs exposing these values in SDK. Changing values would be extremely difficult in SDK if we expose them.

bszekely1: To provide additional context, the device player does not support all mime types per: https://developer.android.com/guide/topics/media/media-formats

We don't want publisher misrepresenting the file types for the player. We selected mp4 since this most demand mime type. Allowing a publisher to set this can cause ads not to render.

Example/PrebidDemoJava/src/main/java/org/prebid/mobile/app/DemoActivity.java
```diff
@@ -73,37 +81,69 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
-                adUnit = new BannerAdUnit(Constants.PBS_CONFIG_ID_320x50, width, height);
-            } else {
-                adUnit = new BannerAdUnit(Constants.PBS_CONFIG_ID_320x50, width, height);
-            }
 
             enableAdditionalFunctionality(adUnit);
 
             if ("DFP".equals(adServerName)) {
-                createDFPBanner(width, height);
+                setupAndLoadAMBanner(width, height);
```
anwzhang: Can you make it in the app UI that two drop downs video and video interstitial so that anyone builds that app can see the demo immediately without modifying the code?

PR #159, opened 2019-11-18
Support for US Privacy Flag

yoalex5:
1. CCPA was added #158
API:
``` Java
//add
SharedPreferences pref = PreferenceManager.getDefaultSharedPreferences(context);
SharedPreferences.Editor editor = pref.edit();
editor.putString("IABUSPrivacy_String", "1YN");
editor.apply();

//remove
SharedPreferences pref = PreferenceManager.getDefaultSharedPreferences(context);
SharedPreferences.Editor editor = pref.edit();
editor.remove("IABUSPrivacy_String");
editor.apply();
```

OpenRTB request:
``` XML
"regs": {
  "ext": {
   "us_privacy": "1YN"
  }
}
```
2. `setSubjectToGDPR()` can take null value to get an IAB value
``` Java
TargetingParams.setSubjectToGDPR(null);
```
3. refactoring
4. tests were added


PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/StorageUtils.java
```diff
@@ -0,0 +1,145 @@
...
+ */
+
+package org.prebid.mobile;
+
+import android.content.Context;
+import android.content.SharedPreferences;
+import android.preference.PreferenceManager;
+import android.support.annotation.Nullable;
+
+final class StorageUtils {
```
anwzhang: Same as iOS comment, should we call this PrivacyUtils?

bszekely1: if we use this utility class for other storage needs one day we should keep it storageUtils, however if we decide to limit it to just for user privacy, we can change it. I would say lets keep it as is since there may be other use cases in the future to read / write to local storage where this class be be used.

yoalex5: `StorageUtils` is a general utility class to work with persistent storage. It will allow us to expand this class in future without generating a lot of small classes. If we had several big logic entities I would create `PrivacyUtils`, `StorageUtils` etc.

Please keep it as is and we will change it when it will be necessary

PR #198, opened 2020-05-29
[Feature Request] fetchBids with PublisherAdRequest$Builder

timrijckaert:
# Preface

This PR is a proof of concept in combination with a feature request.
I also included some open question about the inner working of this Android library and some architectural choices that were made.

# Current problem

Currently (v1.6) when fetching bids via the `AdUnit#fetchDemand` method it allows for any object to be passed.
This object is then checked in the code to see if it is of a known type.

The current supported types are:

    static final String MOPUB_BANNER_VIEW_CLASS = "com.mopub.mobileads.MoPubView";
    static final String MOPUB_INTERSTITIAL_CLASS = "com.mopub.mobileads.MoPubInterstitial";
    static final String DFP_AD_REQUEST_CLASS = "com.google.android.gms.ads.doubleclick.PublisherAdRequest";

We would like to add one more object namely the `PublisherAdRequest$Builder`. 
 
This allows us to manipulate the request after the specific parameters are added for Prebid.  
The way we implemented it in this PR is a non breaking change, but ideally this should be the only way of fetching bids. 
That is because Google's `PublisherAdRequest$Builder` library offers no convenient `copyFrom` method that can take an already existing `PublisherAdRequest` to a builder and therefore Prebid currently blocks us from adding extra values **AFTER** bids were fetched.

# Extra questions

## Reflection
We were a little beetje stumped to see the usage of reflection for adding simple key values in the custom targeting.
Using reflection on mobile is severely bad for performance and should be avoided at ALL costs.

There is probably a good reason why you are not declaring a dependency on the `com.google.android.gms:play-services-ads` lib.
We would love to hear the reasoning.

## Util#removeUsedCustomTargetingForDFP
This method is removing reserved keys from the custom targeting map.
However the map is only allocated after the initial load.
So the first time this method does not remove anything.
The second time a `fetchDemand` call is triggered the `PublisherAdRequest` is a new object so the custom targeting is not even filled in.
Did we miss something here? If not this method seems useless.

Thanks in advance

From your friends at TAR.GET() @ DPG Media ❤️

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/Util.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/Util.java
@@ -474,7 +474,6 @@ public class Util {
     }
 
     private static void handleDFPBuilderCustomTargetingUpdate(HashMap<String, String> bids, Object adObj) {
-        removeUsedCustomTargetingForDFPBuilder(adObj);
         if (bids != null && !bids.isEmpty()) {
             for (String key : bids.keySet()) {
                 Util.callMethodOnObject(adObj, "addCustomTargeting", key, bids.get(key));
@@ -519,10 +518,6 @@ public class Util {
                 bundle.remove(key);
             }
         }
-    }
-
-    private static void removeUsedCustomTargetingForDFPBuilder(Object adRequestObj) {
-        // TODO: How to remove stuff from the custom targeting map?
     }
 
     static <E, U> void addValue(Map<E, Set<U>> map, E key, U value) {
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/Util.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/Util.java
@@ -24,6 +24,7 @@ import android.support.annotation.CheckResult;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.text.TextUtils;
+import android.util.Log;
 import android.view.View;
 import android.view.ViewGroup;
 import android.webkit.ValueCallback;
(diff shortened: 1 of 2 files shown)


yoalex5:
Hello @timrijckaert, thanks for your PR
Your idea with a `PublisherAdRequest$Builder` is clear and it is a good solution. 
@anwzhang, what do you think ?

> Reflection
We are going to fix it in future 

> Util#removeUsedCustomTargetingForDFP
This function is used in conjunction with AdUnit#setAutoRefreshPeriodMillis()

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/Util.java
```diff
@@ -468,6 +473,14 @@ private static void handleDFPCustomTargetingUpdate(HashMap<String, String> bids,
         }
     }
 
+    private static void handleDFPBuilderCustomTargetingUpdate(HashMap<String, String> bids, Object adObj) {
```
yoalex5, line 476: 1. use `removeUsedCustomTargetingForDFP` to support `AutoRefresh` feature
2. use `addReservedKeys` to support `removeUsedCustomTargetingForDFP`

timrijckaert: egerger

yoalex5: Hello @timrijckaert

Does this comment relevant to #199 ?

timrijckaert: Yes I'm sorry I couldn't comment on the other I got a 401 everytime

yoalex5: thanks

bszekely1:
I am closing this PR because this is addressed in PR #199. This PR was originally meant as a POC anyway. Thank you for your contribution @timrijckaert

PR #207, opened 2020-08-03
Bids Response hook

yoalex5:
[GitHub issue for iOS](https://github.com/prebid/prebid-mobile-ios/issues/303)

A new overloaded `fetchDemand` method was created

``` Java
void fetchDemand(@NonNull final OnCompleteListener2 listener)
```

This function:
 1. does not accept an `AdServerObject`(AdManager or MoPub) as a result it does not modify it under the hood
 2. returns Key/Value `targeting` pairs from `auction` response

Publisher is responsible for passing returned result into `AdServerObject`

A good example is `RewardedVideo` for `MoPub` AdServer. 
Usage example:

``` Java
func loadMPRewardedVideo() {
private void loadMPRewardedVideo() {

    adUnit.fetchDemand(new OnCompleteListener2() {
        @Override
        public void onComplete(ResultCode resultCode, Map<String, String> unmodifiableMap) {
            //copy  of unmodifiableMap. It is an extra step to avoid runtime exception during changing unmodifiableMap
            Map<String, String> kvMap  = new HashMap<>(unmodifiableMap);
            String mpKeywords = Util.convertMapToMoPubKeywords(kvMap);
            MoPubRewardedVideoManager.RequestParameters parameters = new MoPubRewardedVideoManager.RequestParameters(mpKeywords);
            MoPubRewardedVideos.loadRewardedVideo(MP_ADUNITID_REWARDED, parameters, (MediationSettings) null);
        }
    });

}
```

**Please note**:
- `unmodifiableMap` is `UnmodifiableMap` which throws `UnsupportedOperationException` in runtime if you try to modify it.

### DIFF
--- a/Example/PrebidDemoJava/src/androidTest/java/org/prebid/mobile/app/AdManagerComplexTest.java
+++ b/Example/PrebidDemoJava/src/androidTest/java/org/prebid/mobile/app/AdManagerComplexTest.java
@@ -39,11 +39,16 @@ import org.prebid.mobile.BannerAdUnit;
 import org.prebid.mobile.Host;
 import org.prebid.mobile.LogUtil;
 import org.prebid.mobile.OnCompleteListener;
+import org.prebid.mobile.OnCompleteListener2;
 import org.prebid.mobile.PrebidMobile;
 import org.prebid.mobile.ResultCode;
 import org.prebid.mobile.Util;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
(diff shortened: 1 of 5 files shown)


anwzhang (approved):

anwzhang (changes requested):
Why create a new ad listener instead of having two methods in the same listener?

anwzhang:
> Why create a new ad listener instead of having two methods in the same listener?

NVM, your implementation is good.

anwzhang (approved):

PR #211, opened 2020-08-31
Demoapp update

anwzhang:
The goal is to move each combination of setup into its own activity, so that when a Publisher is looking for an example, it can quickly go into one activity without being distracted by other examples.


Example/PrebidDemoJava/src/main/java/org/prebid/mobile/app/CustomApplication.java
```diff
@@ -44,8 +44,6 @@ public void onCreate() {
 
         PrebidMobile.setShareGeoLocation(true);
         PrebidMobile.setApplicationContext(getApplicationContext());
-        WebView obj = new WebView(this);
-        obj.clearCache(true);
```
yoalex5, line 48: why did you remove these lines ?

anwzhang: Why do we have this?

yoalex5: I don't know
You added it as a part of commit
c3b0a76b222894f7941f78cb9102cbead724137d

Usually I use this technic to review all requests and responses(no caching) using Charles

anwzhang: I don't remember why I added this, it seems not meaningful to the demo app.
Do you approve the removing?

yoalex5 (approved):

Issue #244, opened 2021-03-22
ArrayIndexOutOfBoundsException in PrebidServerAdapter.java line 73


aielcbmDev:
**Describe the bug**
ArrayIndexOutOfBoundsException in PrebidServerAdapter.java line 73

**To Reproduce**
Unfortunately we don't have the steps to reproduce the issue; only the crashlog from crashlytics. The issue occurs when updating to [Prebid SDK version 1.10](https://github.com/prebid/prebid-mobile-android/releases/tag/1.10). 

**Expected behavior**
The app does not crash

**Smartphone (please complete the following information):**
DEVICE 1:
- Brand: HUAWEI
- Model: HUAWEI P smart 2019
- Orientation: Portrait
- RAM free: 622.86 MB
- Disk free: 19.61 GB
- OS version: Android 10

DEVICE 2:
- Brand: samsung
- Model: Galaxy S20 5G
- Orientation: Portrait
- RAM free: 3.07 GB
- Disk free: 67.63 GB
- OS version: Android 11

**Additional context**
```
Fatal Exception: java.lang.ArrayIndexOutOfBoundsException: src.length=10 srcPos=1 dst.length=10 dstPos=0 length=-1
       at java.lang.System.arraycopy(System.java)
       at java.util.ArrayList.batchRemove(ArrayList.java:726)
       at java.util.ArrayList.removeAll(ArrayList.java:690)
       at org.prebid.mobile.PrebidServerAdapter.stopRequest(PrebidServerAdapter.java:73)
       at org.prebid.mobile.DemandFetcher$RequestRunnable.cancelRequest(DemandFetcher.java:167)
       at org.prebid.mobile.DemandFetcher.destroy(DemandFetcher.java:120)
       at org.prebid.mobile.DemandFetcher$1.run(DemandFetcher.java:145)
       at android.os.Handler.handleCallback(Handler.java:888)
       at android.os.Handler.dispatchMessage(Handler.java:100)
       at android.os.Looper.loop(Looper.java:213)
       at android.app.ActivityThread.main(ActivityThread.java:8178)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:513)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1101)
```

yoalex5:
Hello @aielcbmDev

[...]

Issue #243, opened 2021-03-22
ArrayIndexOutOfBoundsException in PrebidServerAdapter.java line 233


aielcbmDev:
**Describe the bug**
ArrayIndexOutOfBoundsException in PrebidServerAdapter.java line 233

**To Reproduce**
Unfortunately we don't have the steps to reproduce the issue; only the crashlog from crashlytics. The issue occurs when updating to [Prebid SDK version 1.10](https://github.com/prebid/prebid-mobile-android/releases/tag/1.10). 

**Expected behavior**
The app does not crash

**Smartphone (please complete the following information):**
DEVICE 1:
- Brand: samsung
- Model: Galaxy S8
- Orientation: Portrait
- RAM free: 1017.75 MB
- Disk free: 6.69 GB
- OS version: Android 9

DEVICE 2:
- Brand: samsung
- Model: Galaxy A51
- Orientation: Portrait
- RAM free: 775.22 MB
- Disk free: 87.75 GB
- OS version: Android 11

**Additional context**
```
Fatal Exception: java.lang.ArrayIndexOutOfBoundsException: length=10; index=-1
       at java.util.ArrayList.fastRemove(ArrayList.java:557)
       at java.util.ArrayList.remove(ArrayList.java:540)
       at org.prebid.mobile.PrebidServerAdapter$ServerConnector.removeThisTask(PrebidServerAdapter.java:233)
       at org.prebid.mobile.PrebidServerAdapter$ServerConnector.processResult(PrebidServerAdapter.java:214)
       at org.prebid.mobile.PrebidServerAdapter$ServerConnector.onPostExecute(PrebidServerAdapter.java:103)
       at org.prebid.mobile.http.HTTPPost.execute(HTTPPost.java:67)
       at org.prebid.mobile.PrebidServerAdapter$ServerConnector.execute(PrebidServerAdapter.java:109)
       at org.prebid.mobile.PrebidServerAdapter.requestDemand(PrebidServerAdapter.java:61)
       at org.prebid.mobile.DemandFetcher$RequestRunnable$1.run(DemandFetcher.java:188)
       at android.os.Handler.handleCallback(Handler.java:938)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loop(Looper.java:246)
       at android.os.HandlerThread.run(HandlerThread.java:67)
```

PR #231, opened 2021-01-05
Removed AsyncTasks, replaced with executor implementation

avohraa:
* Added package `tasksmanager`, that consists of executors for BG tasks
* Removed AsyncTask implementation throughout the SDK

PR #245, opened 2021-03-23
PrebidServerAdapter errors

yoalex5:
Based on several crash reports #243, #244, #246, #247 this PR was created

I could not reproduce these force closes but all issues has the same root - `serverConnectors` field which throws `ArrayIndexOutOfBoundsException` during remove(`serverConnectors.remove()` and `serverConnectors.removeAll()`)

It seems that several  threads access to `serverConnectors` and modify  it. I propose to use synchronized collection(a lot of list modifications) to protect this behavior

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidServerAdapter.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidServerAdapter.java
@@ -40,6 +40,7 @@ import org.prebid.mobile.http.TaskResult;
 
 import java.lang.ref.WeakReference;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
@@ -48,10 +49,10 @@ import java.util.Map;
 import java.util.UUID;
 
 class PrebidServerAdapter implements DemandAdapter {
-    private ArrayList<ServerConnector> serverConnectors;
+    private final List<ServerConnector> serverConnectors;
 
     PrebidServerAdapter() {
-        serverConnectors = new ArrayList<>();
+        serverConnectors = Collections.synchronizedList(new ArrayList<ServerConnector>());
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


KostiantynTselykh:
@yoalex5 There are 2 more issues, which may relate to concurrency issues fixed in this task: https://github.com/prebid/prebid-mobile-android/issues/248, https://github.com/prebid/prebid-mobile-android/issues/249

KonstantinTselykh:
@yoalex5 this synchronize block might fix the result of the multithreading problem, but I don't think that it will fix the cause. Code invoking `Util.apply` method is marked with `@MainThread` annotation, so if it was really main thread invocation there wouldn't be such concurrent modifications. Example Util.apply invocation from the Prebid SDK source code:
```
@Override
@MainThread
public void onDemandReady(final HashMap<String, String> demand, String auctionId) {
    if (RequestRunnable.this.auctionId.equals(auctionId)) {
        Util.apply(demand, DemandFetcher.this.adObject);
        LogUtil.i("Successfully set the following keywords: " + demand.toString());
        notifyListener(ResultCode.SUCCESS);
    }
}

@Override
@MainThread
public void onDemandFailed(ResultCode resultCode, String auctionId) {
    if (RequestRunnable.this.auctionId.equals(auctionId)) {
        Util.apply(null, DemandFetcher.this.adObject);
        LogUtil.i("Removed all used keywords from the ad object");
        notifyListener(resultCode);
    }
}
```
When debugging I see that the listener method (onDemandReady) is not executed on the main thread so It looks like the annotation is the source of confusion and I'm not sure if there was an intention to invoke listeners from non-main thread. The annotation is incorrect or the thread executing the listeners is wrong?
Screenshot 2021-04-01 at 15 38 34

yoalex5:
These changes were introduced in #231 

@avohraa, what do you think ?

KonstantinTselykh:
@yoalex5 hm, then the whole `PrebidServerAdapter.onPostExecute` (which I think is referring to the AsyncTask.onPostExecute in main thread) is now done on a worker thread and the concurrent collection modification might not  be the only problem here? (if it previously was depending on the main thread execution). E.g. other variables may be affected as well (without crash, but may produce incorrect behaviour when modified and accessed from different threads).

Issue #249, opened 2021-03-31
ConcurrentModificationException in Util line 530


KostiantynTselykh:
**Describe the bug**
ConcurrentModificationException on line 530 in Util

**To Reproduce**
Unfortunately we don't have the steps to reproduce the issue; only the stackstrace.

**Expected behavior**
The app does not crash.

**Additional context**
```
Fatal Exception: java.util.ConcurrentModificationException
java.util.HashMap$HashIterator.nextNode (HashMap.java:1441)
java.util.HashMap$KeyIterator.next (HashMap.java:1465)
org.prebid.mobile.Util.removeUsedCustomTargetingForDFP (Util.java:530)
org.prebid.mobile.Util.handleAdManagerBuilderCustomTargeting (Util.java:463)
org.prebid.mobile.Util.apply (Util.java:415)
org.prebid.mobile.DemandFetcher$RequestRunnable$1$1.onDemandReady (DemandFetcher.java:193)
org.prebid.mobile.PrebidServerAdapter$ServerConnector.notifyDemandReady (PrebidServerAdapter.java:251)
org.prebid.mobile.PrebidServerAdapter$ServerConnector.processResult (PrebidServerAdapter.java:208)
org.prebid.mobile.PrebidServerAdapter$ServerConnector.onPostExecute (PrebidServerAdapter.java:103)
org.prebid.mobile.http.HTTPPost.execute (HTTPPost.java:67)
org.prebid.mobile.PrebidServerAdapter$ServerConnector.execute (PrebidServerAdapter.java:109)
org.prebid.mobile.PrebidServerAdapter.requestDemand (PrebidServerAdapter.java:61)
org.prebid.mobile.DemandFetcher$RequestRunnable$1.run (DemandFetcher.java:188)
android.os.Handler.handleCallback (Handler.java:873)
android.os.Handler.dispatchMessage (Handler.java:99)
android.os.Looper.loop (Looper.java:214)
android.os.HandlerThread.run (HandlerThread.java:65)
```

Issue #248, opened 2021-03-31
NullPointerException in PrebidServerAdapter line 237


KostiantynTselykh:
**Describe the bug**
NullPointerException on line 237 in PrebidServerAdapter.

**To Reproduce**
Unfortunately we don't have the steps to reproduce the issue; only the stackstrace.

**Expected behavior**
The app does not crash.

**Additional context**
 My guess is this place can cause the crash:
`if (connector.getAuctionId().equals(auctionId))` (line 68), but currently not sure how `getAuctionId` can return null.

```
Fatal Exception: java.lang.NullPointerException
Attempt to read from field 'java.lang.String n2.e.a.s$a.d' on a null object reference
org.prebid.mobile.PrebidServerAdapter$ServerConnector.getAuctionId (PrebidServerAdapter.java:237)
org.prebid.mobile.PrebidServerAdapter.stopRequest (PrebidServerAdapter.java:68)
org.prebid.mobile.DemandFetcher$RequestRunnable.cancelRequest (DemandFetcher.java:167)
org.prebid.mobile.DemandFetcher.destroy (DemandFetcher.java:120)
org.prebid.mobile.DemandFetcher$1.run (DemandFetcher.java:145)
android.os.Handler.handleCallback (Handler.java:873)
android.os.Handler.dispatchMessage (Handler.java:99)
android.os.Looper.loop (Looper.java:214)
android.app.ActivityThread.main (ActivityThread.java:7050)
java.lang.reflect.Method.invoke (Method.java)
com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:494)
com.android.internal.os.ZygoteInit.main (ZygoteInit.java:964)
```

yoalex5:
`connector` is null

This behaviour can be reproduced if
```
connector = null;
if (connector.getAuctionId().equals(auctionId)) {
```

@KostiantynTselykh, have you tried #245
Is this issue is reproduced in build with #245 ?

Issue #247, opened 2021-03-23
NullPointerException in PrebidServerAdapter.java line 68


aielcbmDev:
**Describe the bug**
NullPointerException in PrebidServerAdapter.java line 68

**To Reproduce**
Unfortunately we don't have the steps to reproduce the issue; only the crashlog from crashlytics. The issue occurs when updating to [Prebid SDK version 1.10](https://github.com/prebid/prebid-mobile-android/releases/tag/1.10).

**Expected behavior**
The app does not crash

**Smartphone (please complete the following information):**
DEVICE 1:
- Brand: HUAWEI
- Model: Huawei P30
- Orientation: Portrait
- RAM free: 1.92 GB
- Disk free: 65.99 GB
- OS version: Android 10

DEVICE 2:
- Brand: samsung
- Model: Galaxy S9
- Orientation: Portrait
- RAM free: 919.2 MB
- Disk free: 34.55 GB
- OS version: Android 10

**Additional context**
```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String org.prebid.mobile.PrebidServerAdapter$ServerConnector.getAuctionId()' on a null object reference
       at org.prebid.mobile.PrebidServerAdapter.stopRequest(PrebidServerAdapter.java:68)
       at org.prebid.mobile.DemandFetcher$RequestRunnable.cancelRequest(DemandFetcher.java:167)
       at org.prebid.mobile.DemandFetcher.destroy(DemandFetcher.java:120)
       at org.prebid.mobile.DemandFetcher$1.run(DemandFetcher.java:145)
       at android.os.Handler.handleCallback(Handler.java:883)
       at android.os.Handler.dispatchMessage(Handler.java:100)
       at android.os.Looper.loop(Looper.java:237)
       at android.app.ActivityThread.main(ActivityThread.java:8125)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:496)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1100)
```

Issue #246, opened 2021-03-23
ConcurrentModificationException in PrebidServerAdapter.java line 67


aielcbmDev:
**Describe the bug**
ConcurrentModificationException in PrebidServerAdapter.java line 67

**To Reproduce**
Unfortunately we don't have the steps to reproduce the issue; only the crashlog from crashlytics. The issue occurs when updating to [Prebid SDK version 1.10](https://github.com/prebid/prebid-mobile-android/releases/tag/1.10).

**Expected behavior**
The app does not crash

**Smartphone (please complete the following information):**
DEVICE 1:
- Brand: HUAWEI
- Model: HUAWEI P smart 2019
- Orientation: Portrait
- RAM free: 600.56 MB
- Disk free: 14.72 GB
- OS version: Android 10

DEVICE 2:
- Brand: samsung
- Model: Galaxy J4+
- Orientation: Portrait
- RAM free: 437.4 MB
- Disk free: 7.65 GB
- OS version: Android 9

**Additional context**
```
Fatal Exception: java.util.ConcurrentModificationException
       at java.util.ArrayList$Itr.next(ArrayList.java:860)
       at org.prebid.mobile.PrebidServerAdapter.stopRequest(PrebidServerAdapter.java:67)
       at org.prebid.mobile.DemandFetcher$RequestRunnable.cancelRequest(DemandFetcher.java:167)
       at org.prebid.mobile.DemandFetcher.destroy(DemandFetcher.java:120)
       at org.prebid.mobile.DemandFetcher$1.run(DemandFetcher.java:145)
       at android.os.Handler.handleCallback(Handler.java:873)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loop(Looper.java:214)
       at android.app.ActivityThread.main(ActivityThread.java:7094)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:494)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:975)
```

PR #253, opened 2021-04-09
PrebidServerAdapter errors alternative solution

KostiantynTselykh:
This PR shows an alternative way to fix the concurrency issue mentioned in next issues: https://github.com/prebid/prebid-mobile-android/issues/243, https://github.com/prebid/prebid-mobile-android/issues/244, https://github.com/prebid/prebid-mobile-android/issues/246, https://github.com/prebid/prebid-mobile-android/issues/247, https://github.com/prebid/prebid-mobile-android/issues/248, https://github.com/prebid/prebid-mobile-android/issues/249
Original PR which also has the concurrency fixes: https://github.com/prebid/prebid-mobile-android/pull/245

The reason for this alternative PR is that `HttpPost.onPostExecute` (and all future processing in `PrebidServerAdapter`) was initially executed on the main thread. Some of the methods in `PrebidServerAdapter` are marked with `@MainThread`  annotation, but after debugging you'll see that they are no longer executed on main thread (more detailed description and debugging screenshots can be found in the original PR discussion). 
I believe that the above changes lead to concurrency issues in version `1.10`, because these changes were introduced in the following PR: https://github.com/prebid/prebid-mobile-android/pull/231 (AsyncTask.onPostExecute was previously executed on the main thread). This PR should fix the reason of concurrency issues in PrebidServerAdapter.

CC: @yoalex5 @avohraa @bszekely1 

**Commit details**
fix(HttpPost):
- Posting result from HTTPPost on MainThread
- Marked onPostExecute with main thread annotation
- Removed unused methods

test(PrebidNativeNativeTest):
- Applied changed to execute mainThread executor tasks instantly

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidServerAdapter.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidServerAdapter.java
@@ -98,13 +98,13 @@ class PrebidServerAdapter implements DemandAdapter {
             adType = requestParams.getAdType();
         }
 
+        @MainThread
         @Override
         protected void onPostExecute(TaskResult<JSONObject> response) {
             processResult(response);
-
         }
 
-         public void execute() {
+        public void execute() {
             timeoutCountDownTimer.start();
             super.execute();
         }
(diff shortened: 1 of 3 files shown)


YuriyVelichkoOpenX:
@bszekely1, @yoalex5 @avohraa   how we can speed up the review? Publishers are still experiencing crashes and they ask when it is fixed.

yoalex5 (approved):

PR #259, opened 2021-04-29
Rendering feature

KostiantynTselykh:
### Description

This PR includes 6 modules:
- **omsdk-android** - OpenMeasurement SDK library;
- **PrebidMobile-rendering** - Rendering feature and unit tests;
- **PrebidMobile-gamEventHandlers** - Rendering event handlers for integrating with GAM as primary ad server and unit tests.
- **PrebidMobile-mopubAdapters** - MoPub mediation adapters;
- **PrebidInternalTestApp** - Internal rendering demo application and UI tests;
- **TestUtils** - Reflection and utils for testing.

**Build and deployment of rendering modules**
`buildPrebidMobileRendering.sh` script was introduced in the `scripts` folder. The result of this script are rendering module artifacts (.aar) and generated .pom files which are stored in the `generated/rendering/%artifactId%` folder. 
**NOTE:** No changes were made to the `buildPrebidMobile.sh` script.

**Changes in deployPrebidMobile.sh**
- `buildPrebidMobileRendering.sh` script execution step was added to generate the required artifacts. 
- Moved maven deploy command into a separate function with params and executing it for rendering and prebid mobile modules separately.

If you wish to disable rendering module build and deployment you can comment out / remove next lines from `deployPrebidMobile.sh`:

`sh ./buildPrebidMobileRendering.sh`

And

```
renderingArtifacts=("prebid-mobile-sdk-rendering" "prebid-mobile-sdk-mopubAdapters" "prebid-mobile-sdk-gamEventHandlers" "omsdk-android")

for n in "${!renderingArtifacts[@]}"; do
  artifact=${renderingArtifacts[$n]}
  artifactPath="${DEPLOY_DIR_ABSOLUTE}/rendering/${artifact}/${artifact}"

  # no javadoc and sources for omsdk-android artifacts
  if [ "${artifact}" == "omsdk-android" ]; then
    mavenDeploy $"${artifactPath}.pom" $"${artifactPath}.aar"
  else
    mavenDeploy $"${artifactPath}.pom" $"${artifactPath}.aar" $"${artifactPath}-sources.jar" $"${artifactPath}-javadoc.jar"
  fi
done
```

**NOTE:** Deployment of rendering artifacts was not tested as it requires actual deployment to maven repository.

### Changes in Prebid modules

- Updated gradle plugin to v 4.1.3 (in the project `build.gradle` file)
**Reason:** Previous (3.5.0) version of gradle plugin caused the Robolectric library to behave incorrectly and use deprecated LEGACY android resources instead of BINARY resources. This caused unit tests to fail in our rendering module.
**How does it affect prebid modules:** core SDK module was not affected at all. The DrPrebid module requires one minor change which doesn’t affect anything.

- Updated compile SDK of prebid demo modules to 30 (demo apps `build.gradle` files)
**Reason:** Our module depends on some constants which are only available in > API 28 (current version for some prebid modules). 
**How does it affect prebid modules:** PrebidDemoKotlin required minor changes in DemoActivity.

- Added `ext` properties to project `build.gradle` file 
**Reason:** Reduce duplication.
**How does it affect prebid modules:** Next properties can be reused: 
```
    artifactGroupId = "org.prebid"
    artifactFolder = "${buildDir}/generated-artifacts"
    omSdkVersion = "1.3.17"
    omSdkModuleName = "omsdk-android"
```


yoalex5:
Hello @KostiantynTselykh 

Thanks for your PR. I have a small question. [GitHub issue](https://github.com/prebid/prebid-mobile-ios/issues/363) has next component diagram



Which I read as 
```
Application -> PrebidSDK Rendering -> GAM Event Handler
```

But PR has next structure
```
PrebidInternalTestApp -> PrebidMobile-gamEventHandlers -> PrebidMobile-rendering
```

yoalex5:
Prebid SDK supports additional next request keys
```
"startdelay"
"adslot"
"storedauctionresponse"
"storedbidresponse"
"atype"
"uids"
``` 

Also rendering module doesn't support Prebid privacy policy logic, which allows to manage these flags on a local Prebid scope
```
"Prebid_COPPA"
"Prebid_GDPR"
"Prebid_GDPR_consent_strings"
"Prebid_GDPR_PurposeConsents"
```

KostiantynTselykh:
@yoalex5 
Thanks for your comments. Will review and add support for necessary request params.
In the meanwhile a followup question:
What is the purpose of having separated keys for Prebid privacy scope? 
```
"Prebid_COPPA"
"Prebid_GDPR"
"Prebid_GDPR_consent_strings"
"Prebid_GDPR_PurposeConsents"
```
I didn’t find any info about the additional keys in the Prebid documentation. Will be thankful if you provide a link to this information.

Another question: Won’t the additional consent keys break the TCF support?
E.g. let’s look at the next code fragment in `TargetingParams`:

```
  @Nullable
    public static Boolean isSubjectToGDPR() {

        Boolean gdprSubject = null;

        try {
            Boolean pbGdpr = StorageUtils.getPbGdprSubject();
            if (pbGdpr != null) {
                gdprSubject = pbGdpr;
            } else {
                Boolean iabGdpr = StorageUtils.getIabGdprSubject();

                if (iabGdpr != null) {
                    gdprSubject = iabGdpr;
                }
            }
        } catch (PbContextNullException ex) {
            LogUtil.e("Targeting", "can not get GDPR Subject", ex);
        }

        return gdprSubject;
    }

```
You first check the preferences for `Prebid_` values and if they are null - you check the `IAB` prefs. So the `Prebid_` values have priority over `IAB`.
But what if the `Prebid_` and `IAB` GDPR consents won’t match? For example, the publisher will once set the `Prebid_` consent over Prebid SDK API and afterwards the publisher might modify shared prefs directly (or with the help of 3rd party lib) if the user consent changes. The publisher is not aware that Prebid SDK uses a separate key to track consent support and therefore we can end up in situations like in the following test:

```
    @Test
    public void testGDPRFlag() {
        PrebidMobile.setApplicationContext(activity.getApplicationContext());
        // Set GDPR to false through Prebid SDK
        TargetingParams.setSubjectToGDPR(false);
        assertEquals(Boolean.FALSE, TargetingParams.isSubjectToGDPR());

        // Change IABConsent through DefaultSharedPreferences
        final SharedPreferences defaultSharedPreferences = PreferenceManager
            .getDefaultSharedPreferences(activity.getApplicationContext());
        defaultSharedPreferences.edit().putString("IABConsent_SubjectToGDPR", "1").commit();

        // Test will fail, because TCF v2 value is not taken into account.
        assertEquals(Boolean.TRUE, TargetingParams.isSubjectToGDPR());
    }
```
CC: @bszekely1

bszekely1:
Prebid SDK support both TCF1 and 2. Each of them should support passing GDPR applies and the consent string either realtime in the Prebid SDK API or within local storage, where the API should take precedence. The Xandr team wrote this logic and will defer to them for further details. 

@ppuviarasu

ppuviarasu:
For the GDPR, the idea is to provide 1. the API for them to set directly 2. SDK should be able to fetch directly from the user defaults.
Since the `Prebid_` is set directly by the publisher it takes precedence over the value set in the userDefaults

KostiantynTselykh:
@ppuviarasu Thanks for your response. 
Can you please clarify why are these values saved under a separate key (and why they don't use the keys defined in TCF)?

YuriyVelichkoOpenX:
@bszekely1 @ppuviarasu @yoalex5 

So far we do not see reasons to allow publishers to manage consent other than the way that is proclaimed in the TCF guides. So the rendering module won't support the custom consent sources in the open beta release. If we get a reasonable request for a support custom source that won't violate TCF policy we will add this functionality in the same day.

yoalex5 (approved):
d

KostiantynTselykh:
NOTE: This PR is related to the following issue: https://github.com/prebid/prebid-mobile-ios/issues/363

PR #260, opened 2021-04-30
Adapt UI tests

yoalex5:


scripts/testPrebidDemo.sh
```diff
@@ -16,5 +16,10 @@ echo -e "✅ PREBID TESTLOG: $@"
 echoX "clean previous build"
 ./gradlew clean
 
+#building
+echoX "start building UI tests"
+./gradlew assembleSourceCodeDebugAndroidTest
+
+# testing
 echoX "start UI tests"
```
Kowshickkarthick, line 24: Commenting this line means Android UI Tests will not run in Travis.
We have tried fixing the UI Tests why not run them in Travis????

scripts/testPrebidDemo.sh
```diff
@@ -16,5 +16,10 @@ echo -e "✅ PREBID TESTLOG: $@"
...
-./gradlew PrebidDemoJava:connectedSourceCodeDebugAndroidTest
```
Kowshickkarthick, line 20: Commenting this line means Android UI Tests will not run in Travis.
We have tried fixing the UI Tests why not run them in Travis????

yoalex5: Check [this line ](https://github.com/prebid/prebid-mobile-android/pull/260/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485L36)

TravisCI doesn't take UI tests into account because a lot of them are failing. You can run them locally

Kowshickkarthick: OK. Yeah I ran the UI tests in my local and some of them are failing.
Lets fix them later and can be included as a part of TravisCI after fixing them

Kowshickkarthick (approved):

PR #267, opened 2021-06-22
Add Start Auto Refresh support

ow-ro:
Prebid documentation mentions startAutoRefresh [here](https://docs.prebid.org/prebid-mobile/pbm-api/android/pbm-adunit-android.html#startautorefresh), however it is not implemented.

We require the ability to start and stop auto refreshing of ads, for cases where the app is backgrounded or in landscape mode and ads are not visible, we do not want to be making empty requests.

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
@@ -67,11 +67,17 @@ public abstract class AdUnit {
         }
     }
 
+    public void startAutoRefresh() {
+        LogUtil.v("Starting auto refresh...");
+        if (fetcher != null) {
+            fetcher.start();
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


YuriyVelichkoOpenX:
Hi @ow-ro !

It is a great intent to introduce the start/stop functionality that is missed somehow.

However, in order to be consistent with Prebid development flow, please 
- create an issue with the description of the error and proposed fix.  
- add unit tests for new methods in the current PR 
- if you are able - provide a similar solution for iOS SDK. If no - let me know and we'll find who can support it on iOS.

Thanks!

YuriyVelichkoOpenX:
hi @ow-ro!

I've introduced the common for both platforms [refresh policy](https://github.com/prebid/prebid-mobile-ios/issues/393).

Does it makes sense for you?

ValentinPostindustria:
The requesting system doesn't use DemandFetcher anymore. And this functionality was added with the new BidLoader.

PR #268, opened 2021-07-26
Introduce consistent refresh policy

KostiantynTselykh:
This PR introduces changes described in the following issue: 
https://github.com/prebid/prebid-mobile-ios/issues/393

chore(build.gradle):
- Updated compile version to 30
- Updated robolectric version to 4.3.1

feat(AdUnit):
- Added ability to resume autorefresh
- Added unit tests

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
@@ -67,11 +67,17 @@ public abstract class AdUnit {
         }
     }
 
+    public void resumeAutoRefresh() {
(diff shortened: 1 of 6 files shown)


KostiantynTselykh:
**NOTE**: previously the `stopAutoRefresh` caused the `fetcher.destroy` and therefore it couldn't be resumed. Now the `stopAutoRefresh` will actually stop it with ability to resume (as mentioned in docs: https://docs.prebid.org/prebid-mobile/pbm-api/android/pbm-adunit-android.html#stopautorefresh). But the publisher should be careful and stop the refresh when it is no longer required and adunit is no longer used. I think we should mention this in the docs to exclude any possible issues.

YuriyVelichkoPI (approved):
LGTM

Issue #280, opened 2021-10-29
[RENDERING] Make Native API from rendering part private #416


YuriyVelichkoPI:
## Motivation

Keep the single implementation for Native ads. 

Right now we have an almost identical API for native ads. The first one is from the original SDK. The second one is from the Rendering codebase.  The rendering's API is more advanced (supports Native Video, supports OM SDK), but right now we do not have time to merge this API gracefully. So we'll make code private and merge API eventually. 

## Objectives

- Mark classes for Native Ads from Rendering codebase as private.
- Comment all test Native test cases in the InternalTestApp. Make sure that the app is compiled.
- Switch off UI tests for native ads.

## Acceptance Criteria

- Prebid SDK provides only original Native API
- InternalTestApp works and doesn't contain cases for Native API

PR #290, opened 2021-11-12
Make private native ad from rendering. 

ValentinPostindustria:
NativeAdUnit from rendering is unavailable for publishers now. Disabled native test cases in InternalTestsApp. Checked and recompiled all modules and apps in the project. #280


YuriyVelichkoPI (approved):
LGTM

Issue #294, opened 2021-11-18
Migrate to AndroidX


0megaD:
**The Problem:**
It is the year 2021 and Google has dropped support for the Support libraries for years.
In our products, the only dependency still using support libraries is Prebid Mobile.
Google has stated that *all* support libraries have AndroidX replacements, and any new projects built default to the androidX implementation.

At this point, it is only Prebid that is hindering our 100% migration towards androidX.

**The Solution:**
I would like for Prebid to migrate to androidX, or to at least offer an alternative that makes use of the androidX dependencies.

**Alternatives considered:**
We've considered not updating our project and keeping Jetifier enabled, but this slows down compilation times.

**Additional context**
A similar issue was posted years ago, but I couldn't comment in there.
Back then the migration was not considered critical enough (especially considering existing apps might still be using the support libraries)
However there should be a moment in time where it is considered reasonable to no longer support these old libraries.

I hope this suggestion will be revisited.

Another thing to note is that the v28.0.0 `support-annotation` library used by Prebid is binary equivalent to the AndroidX v1.0 version of `androidx.annotation:annotation`. As such, migration should be relatively easy

YuriyVelichkoPI:
Hi @0megaD! @ValentinPostindustria has prepared a PR for the migration and I've asked the community about concerns on their side in the slack. If no objections we'll deploy these changes with upcoming release.

PR #298, opened 2021-11-22
Migration of project to AndroidX libraries. 

ValentinPostindustria:
Google moved all support libraries to AndroidX package. So dependencies in PrebidMobile were replaced with new ones. It allows publishers to disable Jetifier and makes compilation faster. #294


YuriyVelichkoPI (approved):
LGTM

Issue #300, opened 2021-11-25
Add ability to setup app content into Rendering API


YuriyVelichkoPI:
## Motivation

Keep the  consistency with the original prebid API

## Objectives

Duplicate functionality and tests introduced in [this PR](https://github.com/prebid/prebid-mobile-android/pull/274) onto rendering API

PR #302, opened 2021-11-26
Added content url field onto rendering API 

ValentinPostindustria:
Added content object to BannerView, MediationBaseAdUnit, BaseInterstitialAdUnit. Corrected tests to work with content url. #300


PrebidMobile/PrebidMobile-rendering/src/main/java/org/prebid/mobile/rendering/bidding/display/MediationBaseAdUnit.java
```diff
@@ -151,6 +150,10 @@ public String getPbAdSlot() {
         return mAdUnitConfig.getPbAdSlot();
     }
 
+    public void setContentObject(ContentObject content) {
```
YuriyVelichkoPI, line 153: Need to rename this method in the same way as in the PR for the original SDK - `addContent `.

YuriyVelichkoPI (changes requested):

YuriyVelichkoPI (approved):
LGTM

Issue #297, opened 2021-11-22
[RENDERING] Exposure the bid request to publishers 


YuriyVelichkoPI:
## Motivation

Provide an ability to get bid data in runtime. 

### Rendering API

- make the property `mBidResponse` in the class `BannerView` public
- make the property `mBidResponse` in the class `BaseInterstitialAdUnit` public 
- make sure that the publisher is able to read the actual value of the `mBidResponse` in the listener `onAdLoaded` for banner, interstitial, and rewarded integrations. 
- add or modify the existing test to check this behavior 

### Mediation API

We won't change the Mediation API so far.  

- make sure that publishers are able to extract needed data from MoPub's object or implement their own mediation delegate in order to get a pure bid response. 

### Native API

Even though we do not support Native Ads in rendering right now we should update them too, in order to keep behavior consistent. 
- add property `bidResponse` into `NativeFetchDemandResult`.
- add param `bidResponse` to the `init` method of `NativeFetchDemandResult` 
- provide `bidResponse` elsewhere it possible to `NativeFetchDemandResult`

PR #304, opened 2021-11-29
Public bid response 

ValentinPostindustria:
Bid response is public in BannerView and BaseInterstitialAdUnit. Added bid response to NativeFetchDemandResult. Corrected and added test cases. #297


YuriyVelichkoPI (approved):
LGTM

PR #310, opened 2021-12-08
Fixed deploy process for rendering

ValentinPostindustria:
Deploy script publishes rendering module as aar. Build script generates and saves aar files too. Fix Sonatype exception about javadoc and sources in OpenMeasument SDK. Fix possible exception in DemandFetcher. Move staging url to global build.gradle.


YuriyVelichkoPI (approved):
LGTM

Issue #291, opened 2021-11-16
PrebidMobile.timeoutMillis setting does not Appear To Work Correctly


kevinvandenbreemen:
**Describe the bug**
Configuring the timeout in millis on PrebidMobile does not result in prebid requests aborting after the timeout period.  Instead our app is hanging until a response comes back.

**To Reproduce**
1. While configuring PrebidMobile, call ```PrebidMobile.setTimeoutMillis(1000L)```
2. Later make a prebid request as in : ```org.prebid.mobile.AdUnit.fetchDemand(....)```

**Expected behavior**
Callback executes in one second

**Smartphone (please complete the following information):**
 - Device: Samsung Galaxy S6 Edge+
 - OS: Android 7.0
- In our own App

**Additional context**
We are calling a server which has been intermittently performing slower lately.  To reproduce this issue consistently we have been using a proxy (Charles) to throttle connections to the server

PR #326, opened 2021-12-14
Fixed timeout problem

ValentinPostindustria:
Added high priority for Demand handler to work in time. Fixed problem with double demand. #291


YuriyVelichkoPI (approved):
LGTM

PR #339, opened 2022-01-17
Introduce AdMob rewarded adapter

ValentinPostindustria:
Create AdMob rewarded adapter. Add examples to Kotlin test app and Internal test app.


YuriyVelichkoPI (approved):
LGTM

PR #350, opened 2022-02-02
Remove range version in readme file

ValentinPostindustria:
Range version can upload unstable releases, it's better to set up the exact version.


YuriyVelichkoPI (approved):
LGTM

PR #353, opened 2022-02-09
AdMob adapters

ValentinPostindustria:
Create Banner, Rewarded, Interstitial, and Native AdMob adapters. Add examples to Kotlin and Internal test apps. 
Use original API for native instead of rendering one.


YuriyVelichkoPI (approved):
LGTM

PR #359, opened 2022-02-11
Remove native ads components from rendering

### LINKED ISSUE #330: Remove Native Ads components from the Rendering API
## Motivation

According to the product requirements for 1.14 we do not need Native functionality from the rendering API. So we should remove it from the codebase. 

## Prerequisites 

1. #331
2. #332
2. #333

## Objectives

Remove the code for Native Ads from the rendering codebase. 

**Note**: it is not clear from the first glance what we should do with each inconsistent case during replacement. So let's discuss and make decisions online during implementation.

ValentinPostindustria:
Remove native components and tests in the rendering part. Remove native examples from the Internal test app that used rendering components. Closes #330.


YuriyVelichkoPI (approved):
LGTM

PR #365, opened 2022-02-21
Migrate banner cases internaltestapp to pbs on aws

MaksPetrukhnoPI:
* Removed mock server utils classes
* Added stored auction responses setup to bundle
* Renamed mock ad unit with pbs


ValentinPostindustria (approved):
LGTM

PR #379, opened 2022-03-18
Apply unified ad units configuration

ValentinPostindustria:
Place native configuration into original configuration. Remove configuration interfaces. Use one AdSize.


YuriyVelichkoPI (approved):
LGTM

PR #381, opened 2022-03-22
Apply bid loader from rendering instead of demand fetcher

ValentinPostindustria:
Apply bid loader. Create methods for the auto-refresh interval in seconds. Add "test" field to request in rendering.

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
@@ -35,15 +35,11 @@ import org.prebid.mobile.units.configuration.AdUnitConfiguration;
 
 import java.util.*;
 
+import static org.prebid.mobile.PrebidMobile.AUTO_REFRESH_DELAY_MAX;
+import static org.prebid.mobile.PrebidMobile.AUTO_REFRESH_DELAY_MIN;
+
 public abstract class AdUnit {
 
-    private static final int MIN_AUTO_REFRESH_PERIOD_MILLIS = 30_000;
-
-    /**
-     * Auto refresh time period in millis. 0 for disabling auto refresh.
-     * Can't be less than MIN_AUTO_REFRESH_PERIOD_MILLIS.
-     */
-    private int periodMillis = 0;
     protected AdUnitConfiguration configuration = new AdUnitConfiguration();
 
     @Nullable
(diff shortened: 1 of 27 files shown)


PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
```diff
@@ -23,64 +23,72 @@
...
-    private DemandFetcher fetcher;
     protected AdUnitConfiguration configuration = new AdUnitConfiguration();
 
-    AdUnit(@NonNull String configId, @NonNull AdType adType) {
+    @Nullable
+    protected BidLoader bidLoader;
+    @Nullable
+    protected Object adObject;
+
+    AdUnit(@NonNull String configId, @NonNull AdUnitConfiguration.AdUnitIdentifierType adType) {
```
YuriyVelichkoPI, line 50: We shouldn't change an Ad Unit's API. Need to return the `AdType`

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/NativeImageAsset.java
```diff
@@ -82,7 +83,7 @@ public void setW(int w) {
         this.w = w;
     }
 
-    private int h = -1;
```
YuriyVelichkoPI, line 85: Why did we change the default value?

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidMobile.java
```diff
@@ -57,18 +57,18 @@
...
      * Used when the refresh interval is not in the AUTO_REFRESH_DELAY_MIN & AUTO_REFRESH_DELAY_MAX range.
      */
-    public static final int AUTO_REFRESH_DELAY_DEFAULT = 60000;
+    public static final int AUTO_REFRESH_DELAY_DEFAULT = 60_000;
 
     /**
-     * Minimum refresh interval allowed. 15 seconds
+     * Minimum refresh interval allowed. 30 seconds
      */
-    public static final int AUTO_REFRESH_DELAY_MIN = 15000;
```
YuriyVelichkoPI, line 71: We discussed that the default refresh interval for the original API is 30 seconds, but this change shows that it is 15.  We should keep the value of the original API.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/RewardedVideoAdUnit.java
```diff
@@ -17,11 +17,12 @@
 package org.prebid.mobile;
 
 import androidx.annotation.NonNull;
+import org.prebid.mobile.units.configuration.AdUnitConfiguration;
 
 public class RewardedVideoAdUnit extends VideoBaseAdUnit {
 
     public RewardedVideoAdUnit(@NonNull String configId) {
-        super(configId, AdType.REWARDED_VIDEO);
```
YuriyVelichkoPI, line 24: This change looks ambiguous since the `VAST` type is not the same as `REWARDED_VIDEO`. As far as I remember for the rewarded video we use the `isOptIn` flag in the AdUnitConfig.

YuriyVelichkoPI (changes requested):

YuriyVelichkoPI (approved):
LGTM

PR #403, opened 2022-03-30
Add max video duration property

### LINKED ISSUE #386: [RENDERING CONTROLS] Introduce the `maxVideoDuration` property
## Motivation

This issue is a part of #385

## Objectives

- Introduce property `maxVideoDuration` in the `InterstitialAdUnit`, `RewardedAdUnit`, and `AdConfiguration`
- This property indicates the maximum available playback time in seconds. 
- If the value in the `<Duration>` tag is bigger than `maxVideoDuration` SDK should fail with `adDidFailToLoad` providing a respective error message. SDK shouldn't load the video file. 
- Default value - `3600 seconds` i.e. by default any ad is valid.
- Add Unit tests
- Add a respective cast to the InternalTestApp

ValentinPostindustria:
Throw an exception in case of the Prebid ad is longer then max value. Closes #386.


YuriyVelichkoPI (approved):
LGTM

PR #417, opened 2022-04-11
Shut down mopub

MaksPetrukhnoPI:
* removed mopub dependencies from demo apps
* removed mopub dependencies from prebid original api
* removed mopub dependencies from build scripts


ValentinPostindustria (approved):
LGTM

PR #432, opened 2022-04-28
Classes movement and some corrections

ValentinPostindustria:
Move main classes for a publisher to the API package. 
Remove copy Ext object that was created while original and rendering merging. Fix release dependencies, and add open measurement and rendering dependencies.


YuriyVelichkoPI (approved):
LGTM

PR #439, opened 2022-05-04
Remove m prefixes in class fields

ValentinPostindustria:
Rendering SDK uses m- prefixes for class fields, but the original doesn't. So we need to unify the code style. Fix tests that use reflection.


YuriyVelichkoPI (approved):
LGTM

PR #440, opened 2022-05-04
add FLAG_ACTIVITY_NEW_TASK for external browser

clho40:
to fix android.util.AndroidRuntimeException: Calling startActivity() from outside of an Activity  context requires the FLAG_ACTIVITY_NEW_TASK flag.


ValentinPostindustria (approved):
LGTM

PR #448, opened 2022-05-10
Make caching of the bid optional in Rendering #442

### LINKED ISSUE #442: Make caching of the bid optional
## Motivation

Rendering API doesn't require caching of the winning bid. However, it still sends the `cache` property in order to be compatible with the original setup. 

We should introduce an ability to make caching optional if the publisher does not use PBC for analytics or some other purpose. 

## Objectives

- Introduce property `useCacheForReportingWithRenderingAPI` in the `PrebidMobile` class.
- Set the default value to `false`. 
- Add a property description "_Indicates whether the PBS should cache the bid for the rendering API. If the value is true the SDK will make the cache request in order to report the impression event respectively to the legacy analytic setup_."
- If the property is **true** send the `request.ext.prebid.cache`. Otherwise do not send this field in the request. 
- If the property is **true** make sure that SDK makes a dummy cache request alongside the impression event. 
- If the property is **false** the criteria for the winning bid should avoid the `hb_cache_id` key. (According to the doc the winning bid is marked with the next keywords - `hb_bidder`, `hb_size`, and `hb_pb` ). Make sure that the bid response doesn't contain cache keys and endpoints.
- Ignore this flag for the original API.

ValentinPostindustria:
Create useCacheForReportingWithRenderingApi field in PrebidMobile. Add "hb_size" field check for winning bid. Closes #442.


YuriyVelichkoPI (approved):
LGTM

PR #449, opened 2022-05-11
Make Custom headers working again

ibb-jile:
We miss the custom header feature, which we implemented in version 1 (was in HTTPPost class, which is not currently using). So we added it into new version.

previous PR https://github.com/prebid/prebid-mobile-android/pull/337


ValentinPostindustria (approved):
LGTM

PR #455, opened 2022-05-18
Initialization improvements

### LINKED ISSUE #447: Initialization improvements
## Objectives

- Deprecate the `setApplicationContext` method
- Introduce the `initializeSDK()` method with params:
  - `@NonNull Context context`
  -  `@Nullable SdkInitListener listener`
- The listener should provide respective methods for successed and failed initialization:
   - `onSDKInit()`
   - `onSdkFailedToInit(Error error)`
- initializeSDK() should check the status of the Prebid Server via the `status` [endpoint](https://docs.prebid.org/prebid-server/endpoints/pbs-endpoint-status.html). 
- The `onSDKInit()` should be called if:
   - the PBS host and URL are configured properly
   - the call to the `status` endpoint returns `ok`.
- Otherwise, the `onSDKFailedToInit` should be called with providing the respective error message.
- Get and save application context if activity context was placed into the method. 
- Check if previous context weak reference is not null. 
- Replace old `setApplicationContext` calls with new one.

ValentinPostindustria:
Save application context even if activity context was placed. Add request to the status endpoint when init SDK. Check if the previous init context is null when calling init SDK, to have the ability to make reinitialization. Replace all setApplicationContext SDK calls with new ones. Closes #447.

Other: Move all initialization methods to SdkInitializer. Refactor PrebidMobile. Create a new SDK initialization listener. Deprecate the LogLevel field and create a setter.

### DIFF
--- a/Example/PrebidDemoJava/src/main/java/org/prebid/mobile/javademo/CustomApplication.java
+++ b/Example/PrebidDemoJava/src/main/java/org/prebid/mobile/javademo/CustomApplication.java
@@ -19,6 +19,7 @@ package org.prebid.mobile.javademo;
 import android.app.Application;
 import android.os.Build;
 import android.webkit.WebView;
+import org.prebid.mobile.Host;
 import org.prebid.mobile.PrebidMobile;
 import org.prebid.mobile.javademo.utils.ScreenUtils;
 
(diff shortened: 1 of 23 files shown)


YuriyVelichkoPI (approved):
LGTM

PR #477, opened 2022-06-28
Populate device.geo.country property

GenarC:
Populate country property from telephonyManager-locale-reverseGeocoding


ValentinPostindustria (approved):
LGTM

PR #480, opened 2022-07-11
Use OMID Partner name and version from TargetingParams

ibb-jile:
In latest version is not possible overwrite OMID Partner name and version... Method for set is still there (in Targetingparams), but value was not used in final ad request.


ValentinPostindustria (approved):
LGTM.

PR #482, opened 2022-07-20
Update exoplayer to version 2.15.1 and replace deprecated code

gerzalez:
The current version of exoplayer (2.13.3) is incompatible with transitive dependencies brought by other libraries due to exoplayer's removal of deprecated code

Updating exoplayer to at least version 2.15.1 would fix these issues


ValentinPostindustria (approved):
LGTM

PR #486, opened 2022-08-10
fix(Render): Add get query params on url request builder

gerzalez-etermax:
There is an issue when the `vastadtaguri` element in the VAST XML contains an URL with query params.


ValentinPostindustria (approved):
LGTM.

PR #487, opened 2022-08-10
fix(Render): fix NPE in ad response parser vast on get vast url method

gerzalez-etermax:
We found a NPE crash related to empty VAST XML like this:

`<VAST version="2.0"></VAST>`


ValentinPostindustria (approved):
LGTM

PR #491, opened 2022-08-16
fix(Render): replace vast regex validator to include other valid formats

gerzalez-etermax:
Replace vast regex validator to include other valid formats and unified validation criteria on vast parser extractor


ValentinPostindustria:
It's already merged with another pull request.

PR #493, opened 2022-08-22
added OAID logics

clho40:
Included oaid checking and getting, similar to the implementation of gaid

### DIFF
--- a/Example/PrebidInternalTestApp/src/main/java/org/prebid/mobile/renderingtestapp/plugplay/bidding/ppm/PpmVideoFragment.kt
+++ b/Example/PrebidInternalTestApp/src/main/java/org/prebid/mobile/renderingtestapp/plugplay/bidding/ppm/PpmVideoFragment.kt
@@ -19,9 +19,11 @@ package org.prebid.mobile.renderingtestapp.plugplay.bidding.ppm
 import android.os.Bundle
 import android.view.View
 import org.prebid.mobile.AdSize
+import org.prebid.mobile.LogUtil
 import org.prebid.mobile.api.data.VideoPlacementType
 import org.prebid.mobile.api.exceptions.AdException
 import org.prebid.mobile.api.rendering.BannerView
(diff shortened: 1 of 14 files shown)


YuriyVelichkoPI:
Hi @clho404 ! 

The changes look reasonable, however, I wouldn't like to add the HMS dependency to the SDK. 

[AppsFlyer](https://support.appsflyer.com/hc/en-us/articles/360006278797-Android-OAID-implementation-in-the-SDK) and [AdJust](https://help.adjust.com/en/article/oaid-plugin-integration) support OAID via additional dependency. And it looks like a proper way. 

Could you rebuild your solution in order to add a new dependency like PrebidMobile-oaid?

clho40:
@YuriyVelichkoPI that looks like an elegant solution! But how can we do that? How do we create a dependency as a "plugin" to the core Prebid Mobile SDK, and gets invoked when an ad request is being created?

YuriyVelichkoPI:
@clho40, we can add the dependency to the project in the same way as PrebidMobile-admobAdap is implemented. 

Another, and more interesting question, is how to integrate core and oaid modules seamlessly. There is a way to declare an interface for a generic id provider and register libraries that implement it. But I would like to investigate it more profound first. 

Please give me a couple of days since we are now focused on updating the GMA SDK dependency. And after that, I'll return to this question.

As an easier alternative, we can just implement a method in the Targeting class which will set up the oaid. So publishers can extract it on their end, and Prebid still won't depend on HMS.

clho40:
> 

Hello @YuriyVelichkoPI , thanks again for your info.

I am more than happy to get involved in your investigation. If there's an interface for generic id providers, that's perfect! You can also give us some samples / examples of how it should be done. That said, we will check the PrebidMobile-admobAdapter.

The alternative - we are already doing that. We always ask our publishers to manually add the method but it's best if we can make it effortless :)

YuriyVelichkoPI:
related feature request: 
https://github.com/prebid/prebid-mobile-ios/issues/461

PR #508, opened 2022-11-02
User consent manager crash fix

MaksPetrukhnoPI:


PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/sdk/deviceData/managers/UserConsentManager.java
```diff
@@ -80,76 +98,71 @@ public void init(Context context) {
...
-            case GDPR_2_PURPOSE_CONSENT:
-                gdpr2PurposeConsent = preferences.getString(GDPR_2_PURPOSE_CONSENT, null);
-            case COPPA_SUBJECT_CUSTOM_KEY:
-                if (sharedPreferences.contains(COPPA_SUBJECT_CUSTOM_KEY)) {
-                    isSubjectToCoppa = sharedPreferences.getBoolean(COPPA_SUBJECT_CUSTOM_KEY, false);
-                } else {
-                    isSubjectToCoppa = null;
-                }
+        try {
+            switch (key) {
```
ValentinPostindustria, line 124: If you use `switch`, don't forget to add `break` at the end of each case. Now it doesn't work well.

ValentinPostindustria (changes requested):

ValentinPostindustria (approved):
LGTM

PR #510, opened 2022-11-03
Update GAM and OMSDK

ValentinPostindustria:
Always use the latest version of GMA SDK. Create a unit test for checking if GMA was updated. Rewrite AdMob adapters using new API. Update OMSDK. Add a warning about the GMA SDK version to `initializeSDK` method.  

PR requires changes in docs: 
```
val request = AdRequest
    .Builder()
    
    // Old
    .addCustomEventExtrasBundle(PrebidInterstitialAdapter::class.java, extras)
    // New
    .addNetworkExtrasBundle(PrebidInterstitialAdapter::class.java, extras)
```


PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/sdk/SdkInitializer.java
```diff
@@ -92,4 +96,51 @@ public static boolean isIsSdkInitialized() {
...
+
+                    if (prebidVersionBigger) {
+                        LogUtil.info("You should update GMA SDK version to " + PrebidMobile.TESTED_GOOGLE_SDK_VERSION + " version that was tested with Prebid SDK (current version " + googleAdsVersion + ")");
+                    } else if (publisherVersionBigger) {
+                        LogUtil.info("The current version of Prebid SDK is not validated with your version of GMA SDK " + googleAdsVersion + " (Prebid SDK tested on " + PrebidMobile.TESTED_GOOGLE_SDK_VERSION + "). Please update the Prebid SDK or post a ticket on the github.");
+                    }
+                }
+            }
+        } catch (Throwable any) {
+            LogUtil.verbose("Can't get current Google Ads Version!");
```
YuriyVelichkoPI, line 131: We do not need to check GAM for all integration cases. It makes sense only for Original API integration. If SDK is used with MAX such a check and error will be irrelevant. 

Let's implement it like on iOS. We should provide an additional `initializeSdk` method which will have an extra param for `MobileAds` object. The initializer will pass this object into the current method and perform all needed checks. If a publisher doesn't provide the instance of `MobileAds` it won't be an error, and SDK won't generate any false-negative warnings. 

Also, we'll add a description of the new initializer into the docs and ask publishers to use it in case of Original API integration.

YuriyVelichkoPI (changes requested):

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/sdk/SdkInitializer.java
YuriyVelichkoPI: Due to the GMA SDK API there is no way to pass the instance of `com.google.android.gms.ads.MobileAds` into Prebid SDK. So that we should introduce the standalone method 

```
PrebidMobile.checkGoogleMobileAdsCompatibility()
```

Which will perform the validation if the integration scenario requires it.

YuriyVelichkoPI (approved):
LGTM

PR #526, opened 2022-11-22
Provide min width/height percent to Original API

### LINKED ISSUE #524: Provide min width/height in Original API
## Description

Properties `minWidthPerc` and `minHeightPerc` can't be set for Rewarded and Video Interstitial. 

Follow [this comment](https://github.com/prebid/prebid-mobile-ios/issues/701#issuecomment-1320630753) for the details.

ValentinPostindustria:
Closes #524.


YuriyVelichkoPI (changes requested):
We do not have to provide minWidthPerc and minHeightPerc to the video ad units, only for the Display Interstitial. Need to be sure that respective signals are present in the bid request

PR #527, opened 2022-11-24
Fix SERVER_ERROR when NO_BIDS #521

### LINKED ISSUE #521: "No bids" reported as "SERVER_ERROR"
**Describe the bug**
The SDK (the `OnCompleteListener` to be very precise) currently reports `PREBID_SERVER_ERROR` when no bids were found. 
This happens in the case where the `AdUnit` receives an error like this:
```
message: SDK internal error: Invalid bid response: Failed to parse bids. No winning bids were found.
cause: org.prebid.mobile.api.exceptions.AdException: SDK internal error: Invalid bid response: Failed to parse bids. No winning bids were found.
```

I am not 100% sure whether this is a bug since I cannot distinguish between an actual error state and missing bids, but I strongly suspect that in a recent version, the string message was changed and the parsing in `FetchDemandResult` simply gets tripped up. Let me explain why I think this is the case:
1. In `BidResponse` line 152ff the winning bid is simply null-checked, no error, no exception, simply a `null` winning bid
2. The error message is then hardcoded to "Failed to parse bids. No winning bids were found." (unfortunately I cannot trace this string back in history easily)
3. This string is then handed through the layers in the JSON and arrives at the `FetchDemandResult`
4. Here it is compared in line 80ff. with
```
if (msg.contains("No bids")) {
    return NO_BIDS;
}
```
In my opinion this check should succeed and `NO_BIDS` should be reported to the listener / client, so maybe the error message was changed recently and the word "winning" should be removed? 

**To Reproduce**
1. Launch an app with a correct server implementation but no winning bids
2. Check the logs for `AdUnit` (in this case log tag "Prebid") and note the resultCode that is passed to the listener

**Expected behavior**
When there are no bids (but no exceptional error state in the request) the `ResultCode` `NO_BIDS` should be used to allow us developers clean logging and reporting

**Screenshots**
image

**Smartphone**
 - Device: Pixel 6 (probably all Android)
 - OS: Android 13 (probably all Android)
 - SDK Version: 2.0.4

ValentinPostindustria:
Closes #521.

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/data/FetchDemandResult.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/data/FetchDemandResult.java
@@ -77,7 +77,7 @@ public enum FetchDemandResult {
         Matcher interstitialSizeMatcher = invalidInterstitialSize.matcher(msg);
         Matcher impMatcher = storedImpNotFound.matcher(msg);
 
-        if (msg.contains("No bids")) {
+        if (msg.contains("No bids") || msg.contains("No winning bids")) {
             return NO_BIDS;
         }
         if (msg.contains("Timeout")) {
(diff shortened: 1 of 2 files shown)


YuriyVelichkoPI (approved):
LGTM

This is a hotfix change for the release purpose. However, this bug has shown incorrect handling of server responses.  We should fix it in another way which assumes that no bids is not an error is just a response status. Follow the parent ticket for the details.

PR #541, opened 2022-12-06
Consent manager crash fixes

ValentinPostindustria:
Make a request to the `/status` endpoint optional. Add exception handlers for each manager in ManagersResolver.


PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/sdk/StatusRequester.java
```diff
@@ -1,37 +1,37 @@
...
         if (url.contains("/openrtb2/auction")) {
             String statusUrl = url.replace("/openrtb2/auction", "/status");
             ServerConnection.fireWithResult(
                 statusUrl,
-                getResponseHandler(listener)
+                getResponseHandler()
             );
         } else if (url.isEmpty()) {
-            onInitError("Please set host url (PrebidMobile.setPrebidServerHost) and only then run SDK initialization.", listener);
+            LogUtil.error(TAG, "Please set host url (PrebidMobile.setPrebidServerHost) and only then run SDK initialization.");
```
YuriyVelichkoPI, line 28: it's not an error. We just can't recognize the given URL. We shouldn't log it as an error.

PR #553, opened 2022-12-16
Replace GAD application id with new format

ValentinPostindustria:
The Java demo app used the old format of the GAD application id. So after an automatic GAM update, the app throws a runtime exception. 
```
java.lang.RuntimeException: Unable to get provider com.google.android.gms.ads.MobileAdsInitProvider.
```


YuriyVelichkoPI (approved):
LGTM

PR #554, opened 2022-12-19
Remove configuration properties from examples in the Internal app

### LINKED ISSUE #552: Internal Test App - Do not use any custom configs in the cases for Original API
## Motivation

We need to test default configs. For the custom, configs need to provide separate cases or customization options via config page. 

## Objectives

- Make sure that all cases for Original API runs without configuration properties, e,g. `VideoBaseAdUnit.Parameters()` 
- Make sure that we have unit tests for checking the changes in the bi requests due to providing configuration properties.

ValentinPostindustria:
Add unit tests for Video parameters. Closes #552.


YuriyVelichkoPI (approved):
LGTM

PR #555, opened 2022-12-22
Cleanup original bid requests

MaksPetrukhnoPI:


ValentinPostindustria (dismissed):
LGTM

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/VideoAdUnit.java
```diff
@@ -26,6 +27,7 @@ public class VideoAdUnit extends VideoBaseAdUnit {
     public VideoAdUnit(@NonNull String configId, int width, int height) {
         super(configId, AdFormat.VAST);
         configuration.addSize(new AdSize(width, height));
+        configuration.setPlacementType(PlacementType.IN_BANNER);
```
YuriyVelichkoPI, line 30: We shouldn't use IN_BANNER as a default value.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/BasicParameterBuilder.java
```diff
@@ -209,7 +210,12 @@ private void setVideoImpValues(Imp imp) {
 
                 video.minbitrate = videoParameters.getMinBitrate();
                 video.maxbitrate = videoParameters.getMaxBitrate();
-
+                video.linearity = videoParameters.getLinearity();
+                if (videoParameters.getPlacement() != null) {
+                    video.placement = videoParameters.getPlacement().getValue();
+                } else if (adConfiguration.getPlacementTypeValue() != PlacementType.IN_BANNER.getValue()){
+                    video.placement = PlacementType.INTERSTITIAL.getValue();
```
YuriyVelichkoPI, line 217: It looks incorrect. We should set `placement` based on the ad unit type or user-defined property. If user set IN-FEED the SDK should send it.

YuriyVelichkoPI (changes requested):

YuriyVelichkoPI (approved):
LGTM

PR #571, opened 2023-01-27
Remove hb_size from required keywords 

### LINKED ISSUE #567: Do not treat `hb_size` as a marker of a winning bid
## Motivation

The size property doesn't make sense for native ads, and an rtb response doesn't contain such a field. 
Also, looking at the PBS logic, we can see that hb_size can be missed for other ad formats as well. So PBS can skip adding the `hb_size` to the targeting keywords of the winning bid. 

## Objectives

- Remove `hb_size` from the list of markers of the winning bid. 
- Update Unit tests
- Introduce respective ties cases in the Internal test app

ValentinPostindustria:
Closes #567.


YuriyVelichkoPI (approved):
LGTM

PR #577, opened 2023-01-31
Apply network config for Java app

ValentinPostindustria:
The app contains network config, but it's not full and it's not applied to the app.


YuriyVelichkoPI (approved):
LGTM

PR #586, opened 2023-02-07
Changes for /status endpoint

### LINKED ISSUE #499: Update SDK reliance on PBS /status endpoint
**Describe the bug**
The SDK hits the Prebid Server /status endpoint and expects "ok" to be in the response. This is not guaranteed as per the [endpoint doc](https://docs.prebid.org/prebid-server/endpoints/pbs-endpoint-status.html) which says the "exact response can be configured with the status_response config option".

The PBS committee will discuss whether we can provide a query string param that will provide a consistent response. https://github.com/prebid/prebid-server/issues/2431

**To Reproduce**
1. initialize the SDK pointing to any PBS.

To get a sense of the differences, try these:

https://prebid-server.rubiconproject.com/status
https://prebid-server.h5v.eu/status
http://prebid.adnxs.com/pbs/v1/status (note the different path!)
https://prebid.admaxmedia.io/status (endpoint not avail)
https://pbserver.galaxiemedia.fr/status (my favorite results :-)

**Expected behavior**

If a check must be done, it should to be done in a flexible way. e.g. if there's a response, substring search for any of these patterns: "ok", "ready", "good", "active", "up".

Doing a brief scan of known Prebid Servers, some of them move their /status endpoint to a different location. If the SDK wants to hit a status endpoint, then it needs to be flexible in allowing the app to define a PBS-host specific path to the status endpoint.

ValentinPostindustria:
Closes #499.

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidMobile.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidMobile.java
@@ -19,6 +19,7 @@ package org.prebid.mobile;
 import android.content.Context;
 import android.util.Log;
 import android.util.Patterns;
+import android.webkit.URLUtil;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
@@ -206,17 +207,16 @@ public class PrebidMobile {
 
     /**
      * Initializes the main SDK classes and makes request to Prebid server to check its status.
-     * Status request is needed to understand whether server works and responds.
-     * You have to set host url ({@link PrebidMobile#setPrebidServerHost(Host)}) and
-     * /status endpoint url ({@link PrebidMobile#setCustomStatusEndpoint(String)}) before calling this method.
+     * You have to set host url ({@link PrebidMobile#setPrebidServerHost(Host)}) before calling this method.
+     * If you use custom /status endpoint set it with ({@link PrebidMobile#setCustomStatusEndpoint(String)}) before starting initialization.
      * <p>
      * Calls SdkInitializationListener callback with enum initialization status parameter:
      * <p>
-     * SUCCEEDED - initialization and status request were successful.
+     * SUCCEEDED - Prebid SDK is initialized successfully and ready to work.
      * <p>
-     * SERVER_STATUS_WARNING - initialization was successful but the status request was failed.
+     * FAILED - Prebid SDK is failed to initialize and is not able to work.
      * <p>
-     * FAILED - initialization failed.
+     * SERVER_STATUS_WARNING - Prebid SDK failed to check the PBS status. The SDK is initialized and able to work, though.
      * <p>
      * To get the description of the problem you can call {@link InitializationStatus#getDescription()}
      *
@@ -334,12 +334,19 @@ public class PrebidMobile {
      * @see <a href="https://docs.prebid.org/prebid-server/endpoints/pbs-endpoint-status.html">GET /status</a>
      */
     public static void setCustomStatusEndpoint(String url) {
-        if (url != null) {
-            if (url.toLowerCase().startsWith("http") && Patterns.WEB_URL.matcher(url).matches()) {
-                customStatusEndpoint = url;
-            } else {
-                Log.e(TAG, "Can't set custom /status endpoint, it is not valid.");
-            }
+        if (url == null) {
+            return;
+        }
+
+        if (!Patterns.WEB_URL.matcher(url).matches()) {
+            Log.e(TAG, "Can't set custom /status endpoint, it is not valid.");
+            return;
+        }
+
+        if (url.startsWith("http")) {
+            customStatusEndpoint = url;
+        } else {
+            customStatusEndpoint = URLUtil.guessUrl(url).replace("http", "https");
         }
     }
 
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidMobile.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidMobile.java
@@ -17,17 +17,18 @@
 package org.prebid.mobile;
 
 import android.content.Context;
+import android.util.Log;
+import android.util.Patterns;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
-import org.prebid.mobile.api.exceptions.InitError;
+import org.prebid.mobile.api.data.InitializationStatus;
 import org.prebid.mobile.core.BuildConfig;
 import org.prebid.mobile.rendering.listeners.SdkInitializationListener;
 import org.prebid.mobile.rendering.mraid.MraidEnv;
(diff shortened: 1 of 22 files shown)


PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidMobile.java
```diff
@@ -201,12 +205,24 @@ public static HashMap<String, String> getCustomHeaders() {
 
 
     /**
-     * Initializes the main SDK classes. Makes request to Prebid server to check its status.
-     * You have to set host url ({@link PrebidMobile#setPrebidServerHost(Host)}) before calling this method.
+     * Initializes the main SDK classes and makes request to Prebid server to check its status.
+     * Status request is needed to understand whether server works and responds.
+     * You have to set host url ({@link PrebidMobile#setPrebidServerHost(Host)}) and
```
YuriyVelichkoPI, line 210: Let's remove these lines and replace them with 

```
If you use custom /status endpoint set it with ({@link PrebidMobile#setCustomStatusEndpoint(String)}) before starting initialization.
```

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidMobile.java
```diff
@@ -201,12 +205,24 @@ public static HashMap<String, String> getCustomHeaders() {
...
+     * /status endpoint url ({@link PrebidMobile#setCustomStatusEndpoint(String)}) before calling this method.
+     * <p>
+     * Calls SdkInitializationListener callback with enum initialization status parameter:
+     * <p>
+     * SUCCEEDED - initialization and status request were successful.
```
YuriyVelichkoPI, line 215: Lets replace to 
```
SUCCEEDED - Prebid SDK is initializated successfully and ready to work.
```

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidMobile.java
```diff
@@ -201,12 +205,24 @@ public static HashMap<String, String> getCustomHeaders() {
...
+     * <p>
+     * SERVER_STATUS_WARNING - initialization was successful but the status request was failed.
```
YuriyVelichkoPI, line 217: Let's replace with:

```
SERVER_STATUS_WARNING - Prebid SDK failed to check the PBS status. The SDK is initialized and able to work, though. 
```

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidMobile.java
```diff
@@ -201,12 +205,24 @@ public static HashMap<String, String> getCustomHeaders() {
...
+     * <p>
+     * FAILED - initialization failed.
```
YuriyVelichkoPI, line 219: Lets replace to 

```
FAILED - Prebid SDK is failed to initialize and is not able to work.
```

Also need to move this staus to the second position in the description, just after `SUCCEEDED`.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidMobile.java
```diff
@@ -342,6 +327,27 @@ public static void checkGoogleMobileAdsCompatibility(@NonNull String googleAdsVe
...
 
+    /**
+     * Sets full valid URL for the /status endpoint of the PBS.
+     * Request to /status is sent when you call {@link PrebidMobile#initializeSdk(Context, SdkInitializationListener)}.
+     *
+     * @see <a href="https://docs.prebid.org/prebid-server/endpoints/pbs-endpoint-status.html">GET /status</a>
+     */
+    public static void setCustomStatusEndpoint(String url) {
+        if (url != null) {
+            if (url.toLowerCase().startsWith("http") && Patterns.WEB_URL.matcher(url).matches()) {
```
YuriyVelichkoPI, line 338: This check doesn't look accurate. 

`prebid.org` is also valid URL, or `www.prebid.org` or some IP address. But this check won't allow such endpoints. 

Can we use `URLUtil.isValidUrl(url)`?

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidMobile.java
```diff
@@ -342,6 +327,27 @@ public static void checkGoogleMobileAdsCompatibility(@NonNull String googleAdsVe
...
     }
 
+    /**
+     * Sets full valid URL for the /status endpoint of the PBS.
+     * Request to /status is sent when you call {@link PrebidMobile#initializeSdk(Context, SdkInitializationListener)}.
+     *
+     * @see <a href="https://docs.prebid.org/prebid-server/endpoints/pbs-endpoint-status.html">GET /status</a>
+     */
+    public static void setCustomStatusEndpoint(String url) {
+        if (url != null) {
```
YuriyVelichkoPI, line 337: nitpic: I always sugest to use 
- positive logic in the `if` statmets
- make a code as plain as possible
- make an early return wherever it makes sense 

so the following statement is more cleaner 

``` 
void foo(String url) {
    // Process precondition
    if (url == null) {
        return 
    }

    ...

    // Run the function body
}
````

File: PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/BaseNetworkTask.java [hunk 7]
```diff
@@ -52,6 +62,7 @@
     protected static final String CONTENT_TYPE_HEADER = "Content-Type";
     protected static final String CONTENT_TYPE_HEADER_VALUE = "application/json";
 
+    public boolean isStatusRequest = false;
```
YuriyVelichkoPI, line 65: We shouldn't add specific login into the _general_ class. `BaseNetworkTask` shouldn't know the request details.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/sdk/StatusRequester.java
```diff
@@ -26,67 +25,74 @@ public class StatusRequester {
...
-                        if (applicationJson != null) {
-                            onInitializationCompleted();
-                            return;
-                        }
-                    } catch (JSONException exception) {
-                        onInitializationFailed("Wrong `/status` response: " + exception.getMessage());
-                        return;
-                    }
+                int statusCode = response.statusCode;
+                if (statusCode >= 200 && statusCode < 300) {
```
YuriyVelichkoPI, line 55: Can we make this check the part of `BaseNetworkTask`. In order to reuse it in either places like:

```
response.isOkStatusCode
```

YuriyVelichkoPI (changes requested):

YuriyVelichkoPI (approved):
LGTM

PR #589, opened 2023-02-13
Changes for auction request

### LINKED ISSUE #582: SDK First Party support
## Objectives

Look for details at https://github.com/prebid/prebid-mobile-ios/issues/785

ValentinPostindustria:
Closes  #582.

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/AppInfoParameterBuilder.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/AppInfoParameterBuilder.java
@@ -16,6 +16,8 @@
 
 package org.prebid.mobile.rendering.networking.parameters;
 
+import android.text.TextUtils;
+
 import org.prebid.mobile.PrebidMobile;
 import org.prebid.mobile.TargetingParams;
 import org.prebid.mobile.configuration.AdUnitConfiguration;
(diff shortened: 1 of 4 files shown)


YuriyVelichkoPI (changes requested):
Looks good. 

However, it makes sense to rename the `setContext*` methods to `setExt*` since we do not put data into the `context` field anymore. 

let's do it by deprecating of `Context` methods and introducing the `Ext` methods. The `Context` should just call the respective `Ext` method inside. In this way, users will be aware of changes in the method's behavior.

MaksPetrukhnoPI:
Need to add setup first party data arguments from command line in InternalTestApp

Valllent:
@MaksPetrukhnoPI check CL arguments.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/TargetingParams.java
```diff
@@ -493,38 +508,128 @@ public static Map<String, Set<String>> getContextDataDictionary() {
...
+        Set<String> value
+    ) {
+        contextDataDictionary.put(key, value);
+    }
+
+    /**
+     * This method allows to remove specific context data keyword & values set from global context targeting
+     */
+    public static void removeExtData(String key) {
+        contextDataDictionary.remove(key);
```
YuriyVelichkoPI, line 583: need to rename `contextDataDictionary` to `extDataDictionary`

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/AppInfoParameterBuilder.java
```diff
@@ -77,5 +79,10 @@ public void appendBuilderParameters(AdRequestInput adRequestInput) {
         if (!contextDataDictionary.isEmpty()) {
             app.getExt().put("data", Utils.toJson(contextDataDictionary));
         }
+
+        Set<String> contextKeywords = TargetingParams.getContextKeywordsSet();
```
YuriyVelichkoPI, line 83: need to rename and make sure that we don't have deal with a Context

YuriyVelichkoPI (changes requested):
In general looks good. But still, need to rename some internal variables. Also, please, doublecheck the whole codebase one more time by searching the `context` keyword. We shouldn't use it in the scope of requested changes in any way in order not to confuse anyone. Only deprecated methods should contain this word.

YuriyVelichkoPI (approved):
LGTM

PR #592, opened 2023-03-02
Change the bid response parsing in order not to treat No Bids as an error

### LINKED ISSUE #562: [Code Enhancements] Change the bid response parsing in order not to treat No Bids as an error
## Motivation

Keep code and logic clean. 
It's kind of a tech debt which was identified in #521.

## Background 

We treat the absence of a winning bid as a parsing error. 

```
MobileSdkPassThrough bidMobilePassThrough = null;
Bid winningBid = getWinningBid();
if (winningBid == null) {
    hasParseError = true;
    parseError = FetchDemandResult.NO_BIDS_MESSAGE;
    LogUtil.info(TAG, parseError);
}
```

It is not a clear solution because:
- it is not the **parser error**. Everything is fine with parsing here.
- No Bid is a kind of bid response. It is a totally legal state, and it shouldn't be associated with an error. 

## Objectives

- Remove the code that treats the absence of a winning bid as an error. 
- If we should determine the No Bid state of the BidResponse, just check the presence of the result with `getWinningBid()`
- the BidResponse object shouldn't contain any internal errors in the case of No Bid
- update/add unit tests
- User API shouldn't be changed

ValentinPostindustria:
Closes #562. 

Also add null pointer check for Applovin MAX to clean exceptional logs.

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/bidding/data/bid/BidResponse.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/bidding/data/bid/BidResponse.java
@@ -27,7 +27,6 @@ import org.json.JSONException;
 import org.json.JSONObject;
 import org.prebid.mobile.LogUtil;
 import org.prebid.mobile.PrebidMobile;
-import org.prebid.mobile.api.data.FetchDemandResult;
 import org.prebid.mobile.configuration.AdUnitConfiguration;
 import org.prebid.mobile.rendering.models.openrtb.bidRequests.Ext;
 import org.prebid.mobile.rendering.models.openrtb.bidRequests.MobileSdkPassThrough;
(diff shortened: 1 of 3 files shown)


PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/bidding/data/bid/BidResponse.java
```diff
@@ -223,6 +219,10 @@ public boolean isVideo() {
         return false;
     }
 
+    public boolean withNoBids() {
```
YuriyVelichkoPI, line 222: Let's remove this method and encourage the usage of `getWinningBid() == null` if needed. 

In addition, the absence of a winning bid doesn't always mean "no bids". I'm not pretty sure about the PBS behavior, but it may happen that bidder responds with a bid that does meet some requirements, and can't be marked as a winning bid. So the response will contain bids, but no winning bid.

YuriyVelichkoPI (changes requested):

YuriyVelichkoPI (approved):
LGTM

PR #598, opened 2023-03-07
Fix weak reference problem in Requester

### LINKED ISSUE #596: WeakReference in Requester can lead to "withheld" status update
**Describe the bug**
The prebid SDK currently allows for one situation, where a caller is not informed about the result of a `.fetchDemand()` request. This can lead to "infinite loading" and the UI breaking. As far as I can tell, this bug was introduced in 2.x

The class `Requester` houses the important method `makeAdRequest`. Before and after it, there are several well-handled callback updates that cascade through the layers to the caller, for example via `sendAdException` and the `adResponseCallback`. However (I presume since the `UserConsentManager` rewrite for 2.x) there is the case, where if `userConsentManager.canAccessDeviceData()`, the requester making the call is stored in a WeakReference inside the `AdIdInitListener`.

This reference can be cleared by the system if no reference is holding on to it, which will lead the call to `adIdFetchCompletion` or `adIdFetchFailure` (whether successful or not does **not** matter, thus we are actively losing cases in which ads could be shown) to reach line 199, log a warning and be done with it. I have not determined 100% yet what can lead to the reference being cleared, however, I suspect that the dreaded AsyncTask has something to do with it, since the callback is executed in the `FetchAdIdInfoTask:onPostExecute`

**To Reproduce**
1. Add `weakRequester.clear();` before `Requester:L197`
2. Install any of the demo apps on any device that has play services (so that `apiAvailability.isGooglePlayServicesAvailable(context)` in `AdIdManager:L52` succeeds)
3. Observe that there are no success or error callbacks, the UI remains clueless and thinks the request is still going on. This can best be seen in the PrebidInternalTestApp, where neither the "onAdLoaded" nor the "onAdFailed" counters are incremented

**Expected behavior**
There are two levels to this bug. Level 1 is that the problem must be surfaced so that the client can react to the knowledge that no request will be made (for example in the radio.net app we show a loading shimmer during ad load, this shimmer can not be cleared until we know what happened to our fetchDemand). You now might argue that clients should implement a timer to tear down dead requests after some timeout. In that case however, the documentation should communicate this front and center. Also:

Level 2 to solve this problem is finish all work on the Requester instance itself, probably convert it to Kotlin, put the AdId request in a suspend function with an IO-dispatcher (and put it on a scope that is bound to the app, so it survives until the app is dead). The requester shouldn't die in the first place, **I can demonstrate in my app log that this problem can occur when a request was successful and I could totally have called `makeAdRequest` and shown an Ad.** 
I mean... the simplest solution would be for Requester to implement AdIdFetchListener and simply call `AdIdManager.initAdId(context, this);`? But I suppose this would open us up to memory leaks, which is probably why the code is there in the first case? 

I don't know if we should invest more work into making this 100% reproducible (I see the error on average in about 1/7 devices on my desk, and when I had Prebid 2.x in production in the beginning of december last year, that's also roughly how much our ad revenue suddenly dipped) when the volatile state of the instance could be definitively fixed by a small rewrite, when AsyncTask is deprecated anyway? 

**Screenshots**

**Smartphone:**
- Prebid SDK: 2.0.8
- `com.google.android.gms:play-services-ads` version: 21.5.0
- Device / OS: unclear, small RAM sizes might contribute to this? Previously randomly seen on
  - Galaxy S10 Android 12
  - Galaxy S20 Android 13
  - Pixel 1 Android 11

ValentinPostindustria:
Closes #596.

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/modelcontrollers/Requester.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/modelcontrollers/Requester.java
@@ -19,13 +19,25 @@ package org.prebid.mobile.rendering.networking.modelcontrollers;
 import android.content.Context;
 import android.content.res.Resources;
 import android.os.AsyncTask;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
 import org.prebid.mobile.LogUtil;
 import org.prebid.mobile.api.exceptions.AdException;
 import org.prebid.mobile.configuration.AdUnitConfiguration;
 import org.prebid.mobile.rendering.listeners.AdIdFetchListener;
 import org.prebid.mobile.rendering.networking.BaseNetworkTask;
 import org.prebid.mobile.rendering.networking.ResponseHandler;
-import org.prebid.mobile.rendering.networking.parameters.*;
+import org.prebid.mobile.rendering.networking.parameters.AdRequestInput;
+import org.prebid.mobile.rendering.networking.parameters.AppInfoParameterBuilder;
+import org.prebid.mobile.rendering.networking.parameters.BasicParameterBuilder;
+import org.prebid.mobile.rendering.networking.parameters.DeviceInfoParameterBuilder;
+import org.prebid.mobile.rendering.networking.parameters.GeoLocationParameterBuilder;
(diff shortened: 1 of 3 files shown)


YuriyVelichkoPI (approved):
This issue has shown that SDK encourages publishers to use it in a dangerous way that may lead to memory leaks.  The current fix is not super accurate because passing the strong reference to this onto any kind of delayed callback is not the best practice. However we still have to apply it in order to fix regression. 

As a future step, we will perform an SDK architecture audit in the scope of #601.

PR #600, opened 2023-03-15
Applovin MAX fixes

ValentinPostindustria:
Fix Applovin MAX adapter initialization. Add null pointer checkers. Fix not catchable exception in Prebid SDK when called from a background thread.


YuriyVelichkoPI (approved):
LGTM

PR #610, opened 2023-04-07
Multiformat ad units in Original API

### LINKED ISSUE #608: Support multiformat ad units
## Motivation

The initial ticket: https://github.com/prebid/prebid-mobile-ios/issues/809

## Objectives

// TODO

ValentinPostindustria:
Closes #608.


YuriyVelichkoPI (changes requested):
Please, take a look if we can address the migration of instream video.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/data/AdUnitFormat.java
```diff
@@ -16,6 +16,9 @@
 
 package org.prebid.mobile.api.data;
 
+/**
+ * Public ad unit format. Can be set up by a publisher.
+ */
 public enum AdUnitFormat {
     DISPLAY,
```
YuriyVelichkoPI, line 23: Please rename to `BANNER`

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/VideoInterstitialAdUnit.java
```diff
@@ -17,16 +17,24 @@
...
 import org.prebid.mobile.rendering.models.AdPosition;
 import org.prebid.mobile.rendering.models.PlacementType;
 
+import java.util.EnumSet;
+
+/**
+ * @deprecated Use InterstitialAdUnit constructor with adUnitFormats parameter:
+ * `EnumSet.of(AdUnitFormat.VIDEO);`
+ */
+@Deprecated
```
YuriyVelichkoPI, line 31: Can we add a custom message to the compilation warning? It would be great to provide the info about the new approach right in the build log 

image

YuriyVelichkoPI (changes requested):

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/VideoInterstitialAdUnit.java
ValentinPostindustria: We can't add a custom message to the studio. To understand the problem users read the `@deprecated` tag in the documentation or use other static analyzers.

PR #618, opened 2023-04-21
Add generic plugin renderer event listener

github-lucas-bon:
To-do
---

- [x] Resolve conflicts with master
- [x] Remove unnecessary TODO comments
- [x] Add standalone test implementation integrating with a sample plugin listener, such as `Banner 320x50 [Custom Renderer, PluginListener] (In-App)`, same for interstitial
- [x] When checking for compliant plugin renderer, take into account the version as well, currently it checks only the name
- [x] Implement unit tests
- [x] Update the class diagram from the interface `PrebidMobilePluginRenderer` and add it to this pull request

___
What is that
---

This feature add the ability to subscribe to additional ad events from a plugin renderer. 

Currently plugin renderers can only dispatch events from [InterstitialControllerListener](https://github.com/prebid/prebid-mobile-android/blob/feature/add-extra-events-for-plugin-renderer/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/bidding/interfaces/InterstitialControllerListener.java) and [DisplayViewListener](https://github.com/prebid/prebid-mobile-android/blob/feature/add-extra-events-for-plugin-renderer/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/bidding/listeners/DisplayViewListener.java) and more recently from [DisplayVideoListener](https://github.com/prebid/prebid-mobile-android/blob/feature/add-extra-events-for-plugin-renderer/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/bidding/listeners/DisplayVideoListener.java), but external plugins may want to dispatch more events related to the ad life cycle that attend to specific use cases, such as for resizing or any other.

___

How to make it
---

The ideia is to subscribe to an additional event listener from a plugin renderer on the view that is displaying an ad in order to comply with apps that contain complex UI. 

So, for instance, as the publisher can already [set the BannerViewListener](https://github.com/prebid/prebid-mobile-android/blob/5c222b8eec80a5e0e04b707eb926704b366b92a1/Example/PrebidInternalTestApp/src/main/java/org/prebid/mobile/renderingtestapp/plugplay/bidding/ppm/PpmBannerFragment.kt#L68) or a [InterstitialAdUnitListener](https://github.com/prebid/prebid-mobile-android/blob/5c222b8eec80a5e0e04b707eb926704b366b92a1/Example/PrebidInternalTestApp/src/main/java/org/prebid/mobile/renderingtestapp/plugplay/bidding/ppm/PpmInterstitialFragment.kt#L38) when setting the ad on the view, we can as well let they subscribe to an additional plugin event listener to receive callbacks from this specific plugin. 

For example:

```
open class PpmBannerFragment : AdFragment(), BannerViewListener, SampleCustomRendererEventListener 👈👈👈 {

    override fun initAd(): Any? {
        bannerView = BannerView(
            requireContext(),
            configId,
            AdSize(width, height)
        )
        bannerView?.setAutoRefreshDelay(refreshDelay)
        bannerView?.setBannerListener(this)
        bannerView?.setPluginEventListener(this) 👈👈👈
        bannerView?.let { CommandLineArgumentParser.addAdUnitSpecificData(it) }
        binding.viewContainer.addView(bannerView)
        return bannerView
    }
    
    // listen plugin listener callbacks 👈👈👈    
     override fun onRotate() {
         LogUtil.debug("onRotate")
    }
    
    override fun onResizing(height: Float, width: Float) {
          LogUtil.debug("onResizing")
    }
}
```

___

The following diagram shows the implementation for a banner, *but the same logic is followed for insterstitial format.

PrebidMobile — Classes Diagrams (1)

___

PrebidMobile — Classes Diagrams (5)

___

Notice that
---
- Currently the plugin renderer feature is complying only with [Banner/Display and Interstitial](https://docs.prebid.org/formats/formats.html)


YuriyVelichkoPI:
Hi @github-lucas-bon! Sorry for the long silence. 

In general, the requirement to support custom delegates for custom renderers looks highly relevant and even an essential part of the feature. 

However, the implementation should be more subtle.  

The main concern now is a single global listener for the whole plugin.  Such a solution will be hard to utilize for complex UI if ad views belong to different fragments. Or in the case of using the ad units on different activities. 

Can we try to implement an old-fashioned approach?  Prebid's ad should provide a property, `extarnalListners`, an array of  `PluginExtraEventHandler` (to support any amount of renderers). This array should be passed to the particular render on instantiation. The renderer should iterate over listeners and find the needed one.  And then, the rendering engine will use this listener to send custom signals. 

In this case, each ad unit gets its own listener. And it should reduce implementation efforts from the publisher side.

TBD: To avoid heavy type casting operation, we can introduce a single property in the `PluginExtraEventHandler` f.e. `eventHandlerName`, that can be hardcoded in the plugin's implementation and return the name of the plugin.

YuriyVelichkoPI (approved):
LGTM

YuriyVelichkoPI (approved):
LGTM

PR #622, opened 2023-05-03
Replace deprecated Handler constructor

### LINKED ISSUE #620: Don't use deprecated `new Handler()` constructor (without parameter)
please use the constructor with a `Looper` argument (for instance `new Handler(Looper.getMainLooper())`)!
otherwise there will be a crash when the current Looper has not `prepared()` yet.

ValentinPostindustria:
Closes #620.


YuriyVelichkoPI (approved):
LGTM

PR #679, opened 2023-09-01
Allow users to provide custom logging logic

linean:
**Reasoning:**
The current implementation of Prebid relies on Logcat for logging, which may not be suitable for all scenarios. To address this, I propose to introduce a new feature that allows users to specify custom logging logic.

**Solution:**
The core of this enhancement is the introduction of the `PrebidLogger` abstraction. By default, it utilizes the familiar logcat mechanism for logging, ensuring a seamless transition for existing users. However, the real power of this feature lies in its flexibility. Users can now implement their custom logging logic, tailored to their specific needs. Whether it's directing logs to a file, sending them to a remote server, or performing custom filtering and formatting, this enhancement empowers users to take control of their logging process.

I believe that this enhancement will greatly benefit Prebid users who require tailored logging solutions. It enhances the adaptability and robustness of the Prebid library while maintaining backward compatibility for those who prefer the existing logcat-based logging. I look forward to your feedback and collaboration on this pull request to make Prebid even more versatile and user-friendly.


linean:
I've noticed that argument names are obfuscated in the final library. What's the best way to prevent that?

```
public interface PrebidLogger {
    void println(int var1, String var2, String var3);

    void e(String var1, String var2, Throwable var3);
}
```

PR #685, opened 2023-09-12
Fix renderers key node on bid request

github-lucas-bon:
This pull request is to solve the following issue:

[This pull request](https://github.com/prebid/prebid-mobile-android/pull/648) intended to insert renderers on the following node path from the bid request:
`.ext.prebid.sdk.renderers`

But the implementation did not match the expectation [as you can see here](https://github.com/prebid/prebid-mobile-android/blob/0f5a3d8a9cd36b095da96f9ee0e29007a17e849b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/BasicParameterBuilder.java#L438), so the current state is
`.ext.sdk.renderers`

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/bidding/data/bid/Prebid.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/bidding/data/bid/Prebid.java
@@ -27,13 +27,19 @@ import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.prebid.mobile.AdSize;
+import org.prebid.mobile.LogUtil;
 import org.prebid.mobile.PrebidMobile;
 import org.prebid.mobile.TargetingParams;
+import org.prebid.mobile.api.rendering.pluginrenderer.PrebidMobilePluginRegister;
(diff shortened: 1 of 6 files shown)


YuriyVelichkoPI:
Hi @github-lucas-bon ! 

In the server ticket https://github.com/prebid/prebid-server/issues/2908 the `ext.prebid.sdk.renderers` is discussed. And as for me, it makes sense and even necessary to put this extension into the `prebid` object. 

If there are no other conversations or specs about using `ext.sdk.renderers` I believe you should create the PBS issue to follow the requirements in the original PBS ticket.

github-lucas-bon:
> 

We will clarify that in the ticket. There was a mismatching between what was said and what was done. I keep you posted.

YuriyVelichkoPI:
Sorry, @github-lucas-bon, but I still do not understand. 

In the pull request for [PBS GO](https://github.com/prebid/prebid-server/pull/2972), the path is `ext.prebid.sdk.renderers`. As it discussed in the https://github.com/prebid/prebid-server/issues/2908. 

So if the changes for PBS JAVA assume a different path - `ext.sdk.renderers`, it looks inconsistent. Both should follow the same spec. 

What I'm missing here?

YuriyVelichkoPI (dismissed):
LGTM

ChrisHuie (approved):

PR #692, opened 2023-09-19
Fix for 680 NullPointerException in Requester.sendAdException

### LINKED ISSUE #680: NullPointerException in Requester.sendAdException
**Describe the bug**
After recent update from 2.1.0 to 2.1.3 we've started to notice new NullPointerException coming from `Requester.sendAdException`. I hope the following stacktrace will help you to find it.

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke interface method 'void org.prebid.mobile.rendering.networking.ResponseHandler.onErrorWithException(java.lang.Exception, long)' on a null object reference
       at org.prebid.mobile.rendering.networking.modelcontrollers.Requester.sendAdException(Requester.java:152)
       at org.prebid.mobile.rendering.networking.modelcontrollers.Requester.makeAdRequest(Requester.java:169)
       at org.prebid.mobile.rendering.networking.modelcontrollers.Requester$1.adIdFetchCompletion(Requester.java:132)
       at org.prebid.mobile.rendering.utils.helpers.AdIdManager$FetchAdIdInfoTask.onPostExecute(AdIdManager.java:142)
       at org.prebid.mobile.rendering.utils.helpers.AdIdManager$FetchAdIdInfoTask.onPostExecute(AdIdManager.java:100)
       at android.os.AsyncTask.finish(AsyncTask.java:695)
       at android.os.AsyncTask.access$600(AsyncTask.java:180)
       at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:712)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loop(Looper.java:193)
       at android.app.ActivityThread.main(ActivityThread.java:6923)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:870)
 ```

### PR DESCRIPTION
**Pull request by jsligh**
Response Callback could be null so surrounding with Null checks. Closes #680

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/modelcontrollers/Requester.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/modelcontrollers/Requester.java
@@ -88,10 +88,10 @@ public abstract class Requester {
             networkTask.destroy();
         }
         networkTask = null;
-        adResponseCallBack = null;
         if (fetchAdIdInfoTask != null) {
             fetchAdIdInfoTask.cancel(true);
         }
+        adResponseCallBack = null;
         fetchAdIdInfoTask = null;
     }
 

### CODE REVIEW

**Comment by YuriyVelichkoPI** [comment]
@jsligh, please attach the PR to the issues using the following guide
https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

It will help to close the related issues automatically once the PR is merged.

**Review by YuriyVelichkoPI — CHANGES_REQUESTED** [review]
Hi @jsligh, the changes will definitely prevent the crash, but I don't understand the reason why the `adResponseCallBack` can be null here. If you take a look at the flow, you will see that it is pretty straightforward. 

The `load()` method invokes the `sendBidRequest`, which creates the  `BidRequester` with the `responseHandler` as a listener, and, right after that, runs the `startAdRequest`. So the listener should be alive. 

The other question is `getAdId()`, that asynchronously calls the `makeAdRequest()`. But still, the listener is just a local variable of the Requester, so it's extremely strange that it is `null`. 

In this case, it's better to dig deeper and figure out the situation why the `adResponseCallBack` can be null and introduce respective changes. 

It is important not to miss some silent error behavior that will lead to bidding performance issues.

**Comment by YuriyVelichkoPI** [comment]
@linean, please, share the code of Prebid's ad unit integration in your app. I would like to understand the lifetime of the objects due to their scope.

**Comment by linean** [comment]
I cannot really share exact code we use. It's pretty complex and involves different mediations. Simplified version of our Prebid logic would look like that:

```
class PrebidBannerDemo: ComponentActivity() {

    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)

        val maxAdView = MaxAdView("adUnitId", MaxAdFormat.BANNER, this)
        val utils = MaxMediationBannerUtils(maxAdView)
        val adSize = AdSize(320, 50)
        val adUnit = MediationBannerAdUnit(maxAdView.context, "configId", adSize, utils)

        adUnit.fetchDemand {
            maxAdView.loadAd()
        }

        lifecycleScope.launch {
            suspendCancellableCoroutine<Unit> { cancellation ->
                cancellation.invokeOnCancellation {
                    adUnit.destroy()
                    maxAdView.destroy()
                }
            }
        }
    }
}
```

**Comment by linean** [comment]
I should probably mention that crash on our end is pretty rare. It affects less than 0,1% of the users.

**Comment by YuriyVelichkoPI** [comment]
Hi @linean! Thanks for sharing! 

Yes, this `initAdId` adds, I would say, unnecessary async behavior, and it would be great to change it. 

From your code, I see that due to the `lifecycleScope.launch` the `adUnit` wouldn't be destroyed on leaving the scope of `onCreate`, and it looks good. 

So yes, let's take a bit more time to think about improvements around `getAdId` to make initial loading steps synchronous or at least more reliable and straightforward. So, we will definitely know what is happening and why internal objects disappear. 

One more question about the crash info - do you see any common properties among reports, like device model or OS versions?

**Comment by linean** [comment]
These are the results for the past 7 days:

image

As you have control over the `AdIdManager` class, I would suggest incorporating a cancellation feature into it. This would enable the `Requester` to call `adIdManager.destroy()`

**Comment by linean** [comment]
Please keep in mind that even a seemingly small percentage, such as 0.1%, can have a considerable impact. In the case of an application with 10 million DAU, this would translate to 10,000 affected users, which is a significant number. 😢

If I were in charge of the codebase, I would lean towards option 4. Option 3 also looks promising: the `Requester` could maintain a local `destroyed` flag and make decisions based on that.

**Comment by YuriyVelichkoPI** [comment]
We shouldn't tolerate any crashes despite the affected users because they are not our users but users of the publishers who integrate the SDK. Any affected user can write negative reviews of the apps that will harm the publisher's business.

I always vote for stateless (as much as possible) and decoupled solutions. So, making the standalone task that performs a particular job that we can easily cancel looks like the most robust way. However, we need to make sure that it will work fast enough. Could you clarify why the task class should be public in this case?

One more way is to pass the weak reference of the `Requester` to `AdIdFetchListener` so that if the requester is destroyed before the system response, it won't be captured in the listener's methods. As a result - no memory leaks to calls to expired, partially destroyed objects. 

Also, I would like to ask @ValentinPostindustria  to look at our discussion and initial issue because he has already made some fixes and improvements in this class recently, and he can help us not reinvent the old bugs with new fixes :).

File: PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/modelcontrollers/Requester.java [hunk 1]
```diff
@@ -126,25 +125,22 @@ protected void getAdId() {
...
         if (userConsentManager.canAccessDeviceData()) {
             AdIdManager.initAdId(context, new AdIdFetchListener() {
                 @Override
                 public void adIdFetchCompletion() {
                     LogUtil.info(TAG, "Advertising id was received");
-                    if (isDestroyed) {
-                        LogUtil.warning(TAG, "Attempted to make ad request after requester was destroyed");
-                    } else {
-                        makeAdRequest();
+                    if (weakRequester.get() != null) {
```
linean, line 134: This fix won't work. There is no guarantee that garbage collection will run after `Requester` gets destroyed.

YuriyVelichkoPI: Yes, it makes sense. I missed that the `destroy()` method nullifies the `adResponseCallBack`. In this case, introducing the task for obtaining the adId looks more robust and suitable way. It will work in the same way as the `networkTask` in the same class. 

Or will it be overcoming and checking the references in the `sendAdException` enough?

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/modelcontrollers/Requester.java
```diff
@@ -88,6 +89,10 @@ public void destroy() {
         }
         networkTask = null;
         adResponseCallBack = null;
+        if (fetchAdIdInfoTask != null) {
+            fetchAdIdInfoTask.cancel(true);
+        }
+        fetchAdIdInfoTask = null;
```
linean, line 95: ⛏️ That's probably very unlikely to happen but it would be safer first cancel `fetchAdIdInfoTask` and then clear `adResponseCallBack`. Otherwise both can race and still produce the same crash.

YuriyVelichkoPI:
As for me, this implementation is much more reliable and SOLID-like. However, I'm not a big Android expert, so @linean, @ValentinPostindustria, it would be great if you review the changes as well. 

From my side - I want to commit a research design to a different approach for obtaining the advertisement identifier because the current one looks extremely time-consuming. According to the Android [docs](https://developer.android.com/training/articles/ad-id), retrieving the identifier has a 10-second timeout. Due to the comments on our code - we waited 3 seconds. But it is still a big number. 

There are publishers whose app's average screen time is just several seconds. So, to spend so much time on each bid request, it looks like not the optimal approach. 

It would be better to introduce some smart AdIdManager, which will take care of updating the value according to the Android policy but will not block the bid requests. 

If you think it makes sense, I'll prepare the ticket for optimization. And, of course, it is out of the scope of the current ticket. 

cc: @alexsavelyev @mmullin

linean (approved):

linean:
> It would be better to introduce some smart AdIdManager, which will take care of updating the value according to the Android policy but will not block the bid requests.

This is a good idea 👍 It's pretty common among other ad providers to cache this identifier.

YuriyVelichkoPI:
The follow-up task for prefetching the advertising info: 
https://github.com/prebid/prebid-mobile-android/issues/696

YuriyVelichkoPI (approved):
LGTM

PR #694, opened 2023-09-22
Multiformat ad unit

### LINKED ISSUE #688: Multiformat ad unit for Original API
## Objectives

Please use this ticket in the iOS repository as a source of requirements: https://github.com/prebid/prebid-mobile-ios/issues/913

If any Android-specific issues arise let's discuss them here.

ValentinPostindustria:
Closes #688.

### DIFF
--- a/Example/PrebidDemoJava/src/main/java/org/prebid/mobile/javademo/activities/ads/gam/original/GamOriginalApiMultiformatBannerVideoNativeInApp.java
+++ b/Example/PrebidDemoJava/src/main/java/org/prebid/mobile/javademo/activities/ads/gam/original/GamOriginalApiMultiformatBannerVideoNativeInApp.java
@@ -66,16 +66,20 @@ public class GamOriginalApiMultiformatBannerVideoNativeInApp extends BaseAdActiv
     }
 
     private void createAd() {
+        // Random only for test cases. For production use one config id.
         ArrayList<String> configIds = Lists.newArrayList(CONFIG_ID_BANNER, CONFIG_ID_VIDEO, CONFIG_ID_NATIVE);
         String configId = configIds.get((new Random().nextInt(3)));
 
+        // 1. Create PrebidAdUnit with configId
         prebidAdUnit = new PrebidAdUnit(configId);
 
+        // 2. Create PrebidRequest with needed multiformat parameters
         PrebidRequest prebidRequest = new PrebidRequest();
         prebidRequest.setBannerParameters(new BannerParameters());
         prebidRequest.setVideoParameters(new VideoParameters(Lists.newArrayList("video/mp4")));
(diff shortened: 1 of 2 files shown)


PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/original/PrebidAdUnit.java
```diff
@@ -0,0 +1,103 @@
...
+    ) {
+        baseFetchDemand(request, null, listener);
+    }
+
+    public void fetchDemand(
+            PrebidRequest request,
+            Object adObject,
+            OnFetchDemandResult listener
+    ) {
+        baseFetchDemand(request, adObject, listener);
```
YuriyVelichkoPI, line 37: Need to change the order of parameters

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
```diff
@@ -109,7 +124,7 @@ public void fetchDemand(@NonNull final OnCompleteListener2 listener) {
         });
     }
 
-    public void fetchDemand(@NonNull Object adObj, @NonNull OnCompleteListener listener) {
+    public void fetchDemand(@Nullable Object adObj, @NonNull OnCompleteListener listener) {
```
YuriyVelichkoPI, line 127: Why do we change the notation to Nullable? What is the use case? And, can we avoid it? 

It's a public API, and for the publishers, this method should always _demand_ an ad object. 

I see that it's due to migration to BidInfo. However, let's think about how we can avoid changing the public API.

YuriyVelichkoPI (changes requested):

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/data/BidInfo.java
```diff
@@ -0,0 +1,57 @@
...
+import androidx.annotation.Nullable;
+
+import org.prebid.mobile.ResultCode;
+
+import java.util.Map;
+
+public class BidInfo {
+
+    @NonNull
+    private ResultCode resultCode;
```
YuriyVelichkoPI, line 13: Please, rename to `result`

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
ValentinPostindustria: We use this method in the new multiformat API so it's possible that it's null. I removed this annotation and now old users will not think that nullable is okay.

PR #700, opened 2023-10-06
GPID parameter

### LINKED ISSUE #698: Support impression-level GPID parameter
Details in the issue in iOS repo: https://github.com/prebid/prebid-mobile-ios/issues/925

ValentinPostindustria:
Closes #698.


PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
```diff
@@ -406,6 +406,9 @@ public void setPbAdSlot(String pbAdSlot) {
         configuration.setPbAdSlot(pbAdSlot);
     }
 
+    public void setGpid(@Nullable String gpid) {
```
YuriyVelichkoPI, line 409: @ValentinPostindustria  Let's add a getter here as well. I see that there is no bid sense here, but just to be consistent with other properties.

PR #703, opened 2023-10-20
Do not include NativeImageAsset exact dimensions in JSON if not specified

linean:
Based on [OpenRTB Native ads specification](https://www.iab.com/wp-content/uploads/2018/03/OpenRTB-Native-Ads-Specification-Final-1.2.pdf) width and height for image request are optional so that client can only specify minimum required dimensions.

image<br>

In the current implementation this is **not allowed** and width and height are always included in the image request JSON. This bug prevents us from displaying more ads. I would be very grateful for a quick review and including it in the next version.

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/NativeImageAsset.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/NativeImageAsset.java
@@ -20,6 +20,12 @@ public class NativeImageAsset extends NativeAsset {
         hmin = minHeight;
     }
 
+    public NativeImageAsset(int minWidth, int minHeight) {
+        super(REQUEST_ASSET.IMAGE);
+        wmin = minWidth;
+        hmin = minHeight;
+    }
+
     public enum IMAGE_TYPE {
         ICON(1),
         MAIN(3),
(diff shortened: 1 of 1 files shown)


PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/NativeImageAsset.java
```diff
@@ -20,6 +20,12 @@ public NativeImageAsset(int w, int h, int minWidth, int minHeight) {
         hmin = minHeight;
     }
 
+    public NativeImageAsset(int minWidth, int minHeight) {
+        super(REQUEST_ASSET.IMAGE);
+        wmin = minWidth;
```
BaloghTamas, line 25: Shouldn't this one set `v` and `h` to be `-1` for instance, to make sure they are skipped in the json serialization?

linean: `w` and `h` are already initialized as `-1`. You can find them lower in this class.

BaloghTamas: Oh, great 👍

BaloghTamas (approved):

linean:
@YuriyVelichkoPI @jsligh Should I create an issue ticket for this bug?

PR #704, opened 2023-10-31
Added @WorkerThread animation because Requester.Destroy() Requires Background

### LINKED ISSUE #678: NetworkOnMainThreadException when destroying MediationBannerAdUnit
**Describe the bug**
We recently upgraded the Prebid SDK from version 2.1.0 to 2.1.3. After this update, we encountered a new runtime crash scenario when attempting to destroy the `MediationBannerAdUnit`. It appears that this operation is now attempting to perform network operations synchronously, resulting in the Android system throwing a `NetworkOnMainThreadException`.

If `MediationBannerAdUnit.destroy` is not intended to be called from the main thread, I suggest adding an explicit check and update documentation to make it easier for developers to identify and prevent such calls.

**To Reproduce**
This issue is not easily reproducible and requires specific conditions to trigger. Our discovery of this problem was based on analysis of production logs.

**Stacktrace**
```
Caused by android.os.NetworkOnMainThreadException:
       at android.os.StrictMode$AndroidBlockGuardPolicy.onNetwork(StrictMode.java:1273)
       at com.google.android.gms.org.conscrypt.Platform.blockGuardOnNetwork(:com.google.android.gms@233217015@23.32.17 (040300-560149384):5)
       at com.google.android.gms.org.conscrypt.ConscryptFileDescriptorSocket.shutdownAndFreeSslNative(:com.google.android.gms@233217015@23.32.17 (040300-560149384):1)
       at com.google.android.gms.org.conscrypt.ConscryptFileDescriptorSocket.close(:com.google.android.gms@233217015@23.32.17 (040300-560149384):79)
       at com.android.okhttp.Connection.closeIfOwnedBy(Connection.java:1303)
       at com.android.okhttp.OkHttpClient$1.closeIfOwnedBy(OkHttpClient.java:78)
       at com.android.okhttp.internal.http.HttpConnection.closeIfOwnedBy(HttpConnection.java:137)
       at com.android.okhttp.internal.http.HttpTransport.disconnect(HttpTransport.java:319)
       at com.android.okhttp.internal.http.HttpEngine.disconnect(HttpEngine.java:739)
       at com.android.okhttp.internal.huc.HttpURLConnectionImpl.disconnect(HttpURLConnectionImpl.java:126)
       at com.android.okhttp.internal.huc.DelegatingHttpsURLConnection.disconnect(DelegatingHttpsURLConnection.java:93)
       at com.android.okhttp.internal.huc.HttpsURLConnectionImpl.disconnect(HttpsURLConnectionImpl.java:25)
       at com.google.firebase.perf.network.InstrURLConnectionBase.disconnect(InstrURLConnectionBase.java:81)
       at com.google.firebase.perf.network.InstrHttpsURLConnection.disconnect(InstrHttpsURLConnection.java:63)
       at org.prebid.mobile.rendering.networking.BaseNetworkTask.destroy(BaseNetworkTask.java:141)
       at org.prebid.mobile.rendering.networking.modelcontrollers.Requester.destroy(Requester.java:86)
       at org.prebid.mobile.rendering.bidding.loader.BidLoader.destroy(BidLoader.java:180)
       at org.prebid.mobile.api.mediation.MediationBaseAdUnit.destroy(MediationBaseAdUnit.java:275)
       at org.prebid.mobile.api.mediation.MediationBannerAdUnit.destroy(MediationBannerAdUnit.java:60)

jsligh:
Closes #678

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidMobile.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidMobile.java
@@ -92,7 +92,7 @@ public class PrebidMobile {
     /**
      * Tested Google SDK version.
      */
-    public static final String TESTED_GOOGLE_SDK_VERSION = "22.4.0";
+    public static final String TESTED_GOOGLE_SDK_VERSION = "22.5.0";
 
     /**
      * Please use {@link PrebidMobile#setLogLevel(LogLevel)}, this field will become private in next releases.


linean:
I believe the suggested solution is flawed for several reasons:

1. It impacts the public SDK API, necessitating all SDK users to rework their code to invoke the `destroy` method from a background thread. Introducing such breaking changes should allow for a gradual transition.

2. The `MediationBaseAdUnit.destroy` method would deviate from the consistency observed in all other destroy methods within the SDK, such as `MediationNativeAdUnit.destroy`. Addressing this inconsistency would require comprehensive documentation updates, and I am concerned that it might still result in developer errors down the line. It's worth noting that the `WorkerThread` annotation doesn't enforce the use of a background thread; it only issues warnings in certain cases.

3. Moreover, the `MediationBaseAdUnit`, `BidLoader`, `Requester` and other child classes are not properly synchronized, and permitting the `destroy` method to be called from a background thread could potentially give rise to issues like race conditions and memory leaks.

Let's try to prevent introducing numerous new issues while attempting to resolve a single one.

YuriyVelichkoPI:
Hi @linean! Thanks for pointing out the problems, and I'm sorry for the inattentive review. I've missed that it is a part of public API. 

Indeed, this fix shouldn't force publishers to rewrite their code. And there is no logical reason to force destroying the ad unit in the background thread. 

Let's reopen the ticket and think about changing the behavior correctly because this crash can happen with any ad unit due to specific circumstances.

PR #705, opened 2023-11-01
Native event seatbid.bid.ext.prebid.events.imp is not notified

### LINKED ISSUE #686: Native event `seatbid.bid.ext.prebid.events.imp` is not notified
Impression event `seatbid.bid.ext.prebid.events.imp` is not notified when  `adm` doesn't have `eventtrackers`. It must be notified even if there are no `eventtrackers`.

jsligh:
Fixes  #686

### DIFF
--- a/Example/PrebidDemoJava/src/main/java/org/prebid/mobile/javademo/activities/ads/gam/original/GamOriginalApiMultiformatBannerVideoNativeInApp.java
+++ b/Example/PrebidDemoJava/src/main/java/org/prebid/mobile/javademo/activities/ads/gam/original/GamOriginalApiMultiformatBannerVideoNativeInApp.java
@@ -24,6 +24,7 @@ import com.google.android.gms.ads.formats.OnAdManagerAdViewLoadedListener;
 import com.google.android.gms.ads.nativead.NativeAd;
 import com.google.android.gms.ads.nativead.NativeCustomFormatAd;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
 
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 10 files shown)


PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidNativeAd.java
```diff
@@ -309,6 +309,18 @@ public void onImpressionTrackerFired() {
                     });
                     impressionTrackers.add(impressionTracker);
                 }
+            } else if (impEvent != null){
```
YuriyVelichkoPI, line 312: Here are some things that should be improved here:
- we shouldn't use the `else if`. From some perspective, it will fix the issue, but it makes the logic more confusing. The `impEvent` is an independent property that doesn't relate to event trackers. So, it would be incorrect to process them separately only if event trackers are empty. 
- creating the `ImpressionTracker.create(impEvent, ...` we will track the event twice. Once inside the ImoressionTracker, and then in `notifyImpressionEvent()`.
- there is one more registerView method in this class that copies, 😔, the same tracking algorithm and the issue. So we should implement the generic solution for both methods without copying the code.   

I'd suggest to implement it in the following way:
- merge `imp_trackers` and `impEvent` into temp array
- for each entry of the temp array create respective `ImpressionTracker`
- in the `onImpressionTrackerFired` make sure that the `listener.onAdImpression();` will be called only once
- remove `notifyImpressionEvent`, but be sure that the impEvent will be tracked by respective `ImpressionTracker`
- wrap this approach into internal function and call it in both `registerView` methods

YuriyVelichkoPI (approved):
LGTM

PR #709, opened 2023-11-03
Revert "Added @WorkerThread animation because Requester.Destroy() Requires Background"

jsligh:
Reverts prebid/prebid-mobile-android#704


YuriyVelichkoPI (approved):
LGTM

PR #734, opened 2024-01-11
NetworkOnMainThreadException Fix

### LINKED ISSUE #678: NetworkOnMainThreadException when destroying MediationBannerAdUnit

jsligh:
Closes #678


PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/modelcontrollers/Requester.java
```diff
@@ -82,10 +83,32 @@ public abstract class Requester {
...
+     * on a background thread or it throws a NetworkOnMainThreadException
+     **/
+    private static class DestroyNetworkTaskAsyncTask extends  AsyncTask<BaseNetworkTask, Void, Void> {
+
+        @Override
+        protected Void doInBackground(BaseNetworkTask... baseNetworkTasks) {
+            if (baseNetworkTasks.length > 0) {
+                if (baseNetworkTasks[0] != null) {
+                    baseNetworkTasks[0].cancel(true);
+                    baseNetworkTasks[0].destroy();
```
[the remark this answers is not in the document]
YuriyVelichkoPI: And we need to be sure that `BaseNetworkTask` manages the _cancel_ event properly. For all use cases.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/modelcontrollers/Requester.java
```diff
@@ -82,10 +83,31 @@ public abstract class Requester {
 
     public abstract void startAdRequest();
 
+    /**
+     * This Async task is needed because the networkTask needs to be cancelled / destroyed
+     * on a background thread or it throws a NetworkOnMainThreadException
+     **/
+    private static class DestroyNetworkTaskAsyncTask extends  AsyncTask<BaseNetworkTask, Void, Void> {
```
YuriyVelichkoPI, line 90: I meant that `DestroyNetworkTaskAsyncTask` doesn't have much sense since it just helps to make a call to the thread not from the main thread, but still, it is not correct interprocess communication that may lead to unexpected behavior. 

The `networkTask.cancel(true)` should did all the job.

YuriyVelichkoPI: Can we try to remove the DestroyNetworkTaskAsyncTask at all? As for me, it just adds one more multithread call but doesn't influence the general behavior. networks.cancel(true) should stop the thread gracefully without direct communication with the network layer so the error should not appear. 

The error in the initial ticket appears due to the networkTask.destroy(); call wich triggers the network communication on the main thread.

YuriyVelichkoPI (changes requested):
Please review the comment. If we can solve the issue with less code it would be great. If it appears that we still need the extra class - we will keep it. But we at least will try.

PR #736, opened 2024-01-17
Run getGlobalVisibleRect on the main thread instead of a background thread

### LINKED ISSUE #735: Crash for MRAID ads
**Describe the bug**
The bug was reported in the Slack channel: 

I just noticed that the JavaScriptInterface method getCurrentPosition inhence the UI related method View.getGlobalVisibleRect is called in JavaBridge thread, which is a background thread. And this causes a lot of native crashes in our app.

https://github.com/prebid/prebid-mobile-android/blob/master/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/views/webview/mraid/BaseJSInterface.java#L175

**To Reproduce**

No steps. Prod crash.

**Expected behavior**

No crash. 

**Screenshots**

image

image

- Android

jsligh:
Closes #735

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/views/webview/mraid/BaseJSInterface.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/views/webview/mraid/BaseJSInterface.java
@@ -180,6 +180,7 @@ public class BaseJSInterface implements JSInterface {
         RunnableFuture<Void> task = new FutureTask<>(mainThreadRunnable, null);
         try {
             mainHandler.post(task);
+            task.get();
             position.put(JSON_X, (int) (rect.left / Utils.DENSITY));
             position.put(JSON_Y, (int) (rect.top / Utils.DENSITY));
             position.put(JSON_WIDTH, (int) (rect.right / Utils.DENSITY - rect.left / Utils.DENSITY));


PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/views/webview/mraid/BaseJSInterface.java
```diff
@@ -171,9 +171,8 @@ public String getDefaultPosition() {
     public String getCurrentPosition() {
         JSONObject position = new JSONObject();
         Rect rect = new Rect();
-
-        adBaseView.getGlobalVisibleRect(rect);
-
+        Handler mainHandler = new Handler(Looper.getMainLooper());
+        mainHandler.post(() -> adBaseView.getGlobalVisibleRect(rect));
```
YuriyVelichkoPI, line 175: The quick search shows that there are more than 20 usages of the `adBaseView` object in the `BaseJSInterface` class. 

image

I'm afraid any of them can lead to a similar crash.

I'd suggest really, hard to reproduce the current crash first, then fix it, then try to reproduce with other usages of the `adBaseView` before the release. It will help to understand the curcomstanses and prevent other crashes, potentially in the same ads. 

If we can't reproduce it, of course, we need to release the fix based on the assumptions, but it is better to cover as much potential cases as possible. Otherwice we will get the same crash statistic but with another method, for example: ` adBaseView.getGlobalVisibleRect(rect)`

YuriyVelichkoPI (changes requested):

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/views/webview/mraid/BaseJSInterface.java
YuriyVelichkoPI: It looks like the Handler runs the task asynchronously. However, we need the data synchronously to process it in the next step.

image

YuriyVelichkoPI (approved):
Ok, it should fix the current crash. But I'm not sure about all other possible cases. 

Ideally, since we've faced such an issue, we should wrap a managed WebView in some background-thread-safe-invocation wrapper. To make sure that all clients of the JS bridge (ad code and SDK code) can use in safely. It could be done in the scope of the separate tech debt ticket.

YuriyVelichkoPI:
I think we should release this hotfix as soon as possible, since it leads to multiple crashes right now.

PR #750, opened 2024-04-29
refactor: Add nullability annotations

shinwan2:
Add annotations to some symbols to be similar with Kotlin.
Does Prebid have Kotlin integration on its roadmap for the `PrebidMobile-core`?


shinwan2:
@jsligh Understood. Is there any plan for Kotlin integration in `PrebidMobile-core` on `3.0.0`?

PR #754, opened 2024-05-29
Add support for privacy url

andriiginting:
As one of [Native Markup Response Object](https://www.iab.com/wp-content/uploads/2018/03/OpenRTB-Native-Ads-Specification-Final-1.2.pdf), If support is indicated in the request, the privacy URL of the informing page users about buyers targeting activity.


YuriyVelichkoPI (approved):
LGTM

I'd also suggest to align the iOS API. Now the `privacy` is available through `nativeAd.nativeAdMarkup.privacy`, but it will be more consistent if to add a property to the `NativeAd` as we do with other similar native markup data: 

```
    @objc public var eventTrackers: [NativeEventTrackerResponse]? {
        return nativeAdMarkup?.eventtrackers
    }
```

andriiginting:
Hi @jsligh , in which version will this feature request be released?

PR #760, opened 2024-06-04
Return error if FetchDemandResult Contains "no internet"

### LINKED ISSUE #668: Incorrect fetch demand result when internet connection is not available
**Describe the bug**
I have noticed unexpected SDK behavior when I disconnect the device from the network when Prebid ad is auto-refreshing. `MediationBannerAdUnit.fetchDemand` returns `FetchDemandResult.SERVER_ERROR` and I think the correct one should be either `NETWORK_ERROR` or `TIMEOUT`.

I've checked implementation details and `FetchDemandResult` is parsed from the error message which in my case is:
```
SDK internal error: Invalid bid response: Initialization failed: No internet connection detected
```

I propose we change `FetchDemandResult.parseErrorMessage` to check if the message contains `No internet connection detected` and return `NETWORK_ERROR`.

**To Reproduce**
1. Show Prebid banner with auto refresh enabled
2. Disable network connection
3. Observe `FetchDemandResult` from `MediationBannerAdUnit.fetchDemand
4. You get `FetchDemandResult.SERVER_ERROR` suggesting there is some issue with the server

**Expected behavior**
SDK should return `NETWORK_ERROR` because the network is not available.

**Smartphone**
 - SDK Version: 2.1.0

jsligh:
Fixes #668

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/data/FetchDemandResult.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/data/FetchDemandResult.java
@@ -85,7 +85,7 @@ public enum FetchDemandResult {
         if (msg.contains("Timeout")) {
             return TIMEOUT;
         }
-        if (msg.contains("Network Error")) {
+        if (msg.contains("Network Error") || msg.contains("No internet")) {
             return NETWORK_ERROR;
         }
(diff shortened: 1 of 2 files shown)


YuriyVelichkoPI (approved):
LGTM 

I also suggest aligning the iOS behavior for this use case. If the iOS SDK logs a similar misleading message, it is worth fixing.

PR #761, opened 2024-06-05
Set SDK Initialization True before Completion Handler Invocation 

### LINKED ISSUE #711: isSdkInitialized returns false in SdkInitializationListener after succeeded initialization
**Describe the bug**
When calling 
`PrebidMobile.initializeSdk(context) {
    Log.d("Highfivve", "Successfully initialized PrebidMobile: ${it.name}")
    Log.d("Highfivve", "Prebid Mobile initialized: ${PrebidMobile.isSdkInitialized()}")
}`
i would expect that `isSdkInitialized` returns `true`.
This would not be a problem, if I wouldn't want to request bids inside the InitializationListener.
When I try this, this leads to the following error:
`PrebidBidLoader: SDK wasn't initialized. Context is null.`

**To Reproduce**
1. Initialize PrebidMobile
2. Call `PrebidMobile.isSdkInitialized()` inside the initialization listener
3. Request Bid inside the InitializationListener

**Expected behavior**
I would expect `PrebidMobile.isSdkInitialized()` to return `true` and also that bids can be requested inside the InitializationListener.

**Screenshots**
- None

**Desktop (please complete the following information):**
- None

 - Android (using Prebid-Mobile)

**Additional context**
The fix would be to move [this line](https://github.com/prebid/prebid-mobile-android/blob/b73eecd04500d5abb1c7978eb9965e9b4ef6f0eb/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/sdk/InitializationNotifier.java#L57) above the `listener.onInitializationComplete`

I just wanted to discuss whether there is any reason for this behaviour or if I can just submit a merge request to fix this behaviour.

jsligh:
Fixes #711

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/sdk/InitializationNotifier.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/sdk/InitializationNotifier.java
@@ -42,24 +42,25 @@ public class InitializationNotifier {
                     postOnMainThread(() -> {
                         listener.onInitializationComplete(InitializationStatus.SUCCEEDED);
                         listener.onSdkInit();
+                        listener = null;
                     });
                 }
             } else {
                 LogUtil.error(TAG, statusRequesterError);
 
                 if (listener != null) {
-                    InitializationStatus serverStatusWarning = InitializationStatus.SERVER_STATUS_WARNING;
-                    serverStatusWarning.setDescription(statusRequesterError);
                     postOnMainThread(() -> {
+                        InitializationStatus serverStatusWarning = InitializationStatus.SERVER_STATUS_WARNING;
+                        serverStatusWarning.setDescription(statusRequesterError);
                         listener.onInitializationComplete(serverStatusWarning);
                         listener.onSdkFailedToInit(new InitError(statusRequesterError));
+                        listener = null;
                     });
                 }
             }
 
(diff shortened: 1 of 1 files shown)


YuriyVelichkoPI (changes requested):
There is a change in state management due to the misleading name of the flag. 

The `tasksCompletedSuccessfully` indicates that tasks are just completed. SDK doesn't consider the status of tasks.  So we should set  

``` 
tasksCompletedSuccessfully = true; 
```

Otherwise, the function `isSdkInitialized` might return false if the publisher didn't provide a listner. 

In any case, I'd suggest to fix the issue not by moving the place of changing value of the flag but calling the listener asynchronously.

YuriyVelichkoPI:
It shouldn't, if to add the task (call the lister function) to the queue using, for example, `Handler(Looper.getMainLooper())`

YuriyVelichkoPI:
Oh, I missed the point that `postOnMainThread` is our internal function. The name mislead me to https://docs.mapbox.com/android/maps/api/10.18.1/mapbox-maps-android/com.mapbox.maps.threading/-animation-thread-controller/post-on-main-thread.html

Which says: 
> If called from Android Main thread - will be executed immediately in the same callchain otherwise will be posted on Android Main thread explicitly.

But in this case - it is just a wrapper for `Handler(Looper.getMainLooper())`, so it should work great!

YuriyVelichkoPI (approved):
LGTM

Issue #773, opened 2024-06-21
Prebid SDK Android supports to send OpenRTB 'hwv' object


takenorim:
## Type of issue
Unlike the Prebid SDK iOS ([here](https://github.com/prebid/prebid-mobile-ios/issues/1012)), it looks Prebid SDK Android does not support to read and send `hwv`.
- https://github.com/prebid/prebid-mobile-android/blob/027fde2aafb9be39e792169a3be4eaf6c05de883/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/models/openrtb/bidRequests/Device.java#L41-L43
- https://github.com/prebid/prebid-mobile-android/blob/027fde2aafb9be39e792169a3be4eaf6c05de883/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/DeviceInfoParameterBuilder.java#L70-L79

## Description
Unlike iPhone devices, I think the `hwv` of Android devices is not standardized. But, I notice some people tries to read device model via Android SDK. This is worthwhile to consider as Prebid SDK.
- https://stackoverflow.com/questions/1995439/get-android-phone-model-programmatically-how-to-get-device-name-and-model-prog
- https://developer.android.com/reference/android/os/Build

## Goals
Read and send `hwv` to Prebid Server from the Prebid SDK

PR #774, opened 2024-06-21
Introduce `hwv` parameter 

ValentinPostindustria:
Issue #773.


ValentinPostindustria:
It's already covered with the test in `DeviceTest`. This field existed before, but it wasn't set.

Issue #755, opened 2024-05-29
Investigate creating API documentation for Android Prebid SDK


bretg:
We would like to utilize javadocs in Prebid SDK, but running into a problem.

I read various sources for how to get started with javadocs, like https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html

In theory, it's easy, but javadoc is choking on imports:

% javadoc PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidMobile.java
Loading source file PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidMobile.java...
Constructing Javadoc information...
PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidMobile.java:19: error: package android.content does not exist
import android.content.Context;

Searching around for a [solution](https://stackoverflow.com/questions/75246690/exclude-internal-java-classes-when-generating-the-javadoc-of-an-android-library), looks like we might need to use a `gradle` task. Found that `publisher.gradle` already has a `javadoc` task, but trying that path results in an error:

% gradle javadoc
…
Unrecognized VM option 'MaxPermSize=2048M'

This might have to do with my local java environment, which is Corretto 17. (?)

The goal of this issue is to find if there's a way to build javadocs while ignoring import references.

Note that we need to do this from the command line in the long run because we're going to want to script the update of the documentation if it's going to be useful. i.e. I'm not as interested if there's an involved process of setting up an IDE that works. If the process of publishing API reference changes requires a lot of effort, it won't get done and the online docs will fall behind.

YuriyVelichkoPI:
I'm closing this issue since the API docs were properly upgraded during the last release build `2.3.1`.

PR #791, opened 2024-08-06
Generate combined Javadoc documentation

ValentinPostindustria:
[Android issue](https://github.com/prebid/prebid-mobile-android/issues/755). [Original issue](https://github.com/prebid/prebid-mobile-ios/issues/990). 

The URL of the generated documentation: https://docs.prebid.org/prebid-mobile-android/index.html


.github/workflows/upload-docs.yml
```diff
@@ -0,0 +1,48 @@
...
+    - name: Commit documentation changes
+      run: |
+        git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
+        git config --local user.name "github-actions[bot]"
+        git status
+        TAG_NAME=${{ github.ref }}
+        TAG_NAME=${TAG_NAME#refs/tags/}
+        echo "Current tag: $TAG_NAME"
+        ls -l
+        rm -rf docs
```
bretg, line 39: Seems like there ought to be a check here to make sure that docs_new actually exists and is of a certain size before removing the current `docs` dir. e.g. use `du -s docs` and abort the operation if less than some reasonable size.

.github/workflows/upload-docs.yml
```diff
@@ -0,0 +1,48 @@
...
+        cp -r PrebidMobile/build/combinedJavadoc docs_new
+    - name: Checkout repository
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+        ref: docs
+        clean: false
+    - name: Commit documentation changes
+      run: |
+        git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
bretg, line 32: Where did this 41898282+github-actions email come from? I guess it's fine, just not familiar with it.

bretg (changes requested):
Amazing progress. I'd imagine there are probably tweaks we'll want to make to the actual docs, but my only substantive comment is to consider making the github action a little more robust.

.github/workflows/upload-docs.yml
ValentinPostindustria: It's an email for GitHub actions bot. [Ref](https://github.com/marketplace/actions/hexo-action#refer-anchor-1).

bretg (approved):
Thanks for making the workflow more robust. Note that my review is quite limited. @YuriyVelichkoPI will need to sign off with a more robust review.

PR #804, opened 2024-11-14
Impression tracker for original API

ValentinPostindustria:


PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
```diff
@@ -449,6 +477,8 @@ public void onFetchCompleted(BidResponse response) {
                 HashMap<String, String> keywords = response.getTargeting();
                 Util.apply(keywords, adObject);
                 originalListener.onComplete(ResultCode.SUCCESS);
+
+                registerVisibilityTrackerIfNeeded(bidResponse);
```
YuriyVelichkoPI, line 481: This approach doesn't look reliable. Switching the tracker on/off for each result of the bid fetching may lead to gaps in tracking.

Let's implement it in the way introduced on iOS.

- start viewability tracking on calling of `activatePrebidImpressionTracker` 
- stop tracking on destroying the AdUnit object 
- once the tracked view meets the impression tracking criteria - check the content, and if it contains PUC - track the respective impression.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
```diff
@@ -65,7 +72,11 @@ public abstract class AdUnit {
     @Nullable
     protected BidResponse bidResponse;
 
+    protected final VisibilityMonitor visibilityMonitor = new VisibilityMonitor();
+    protected WeakReference<View> adViewReference = new WeakReference<>(null);
```
YuriyVelichkoPI, line 76: Can we hold adViewReference in VisibilityMonitor?

YuriyVelichkoPI (changes requested):

ValentinPostindustria:
[Charles session](https://github.com/user-attachments/files/17835511/burl_tracking.zip).

ValentinPostindustria:
[Charles session with Mraid](https://github.com/user-attachments/files/17844060/burl_tracking_with_mraid.zip).

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
```diff
@@ -494,5 +519,45 @@ public AdUnitConfiguration getConfiguration() {
         return configuration;
     }
 
+    private void registerVisibilityTrackerIfNeeded(BidResponse response) {
+        visibilityMonitor.stopTracking();
```
YuriyVelichkoPI, line 523: Is it safe that we stop tracking when we receive the bid? The SDK can get a new bid before the impression of the previous one?

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
```diff
@@ -494,5 +519,45 @@ public AdUnitConfiguration getConfiguration() {
...
+
+        String burl = response.getWinningBid().getBurl();
+
+        String cacheId = response.getTargeting().get("hb_cache_id");
+        if (cacheId == null) {
+            LogUtil.warning(TAG, "Can't register visibility tracker. There is no hb_cache_id keyword.");
+            return;
+        }
+
+        boolean isBannerTracker = !(this instanceof InterstitialAdUnit);
```
YuriyVelichkoPI, line 541: This code looks confusing. Please make the activation conditions more clear.

Example/PrebidInternalTestApp/src/main/java/org/prebid/mobile/renderingtestapp/plugplay/bidding/gam/original/GamOriginalInstreamNewApiFragment.kt
```diff
@@ -101,7 +101,7 @@ class GamOriginalInstreamNewApiFragment : AdFragment() {
 
     override fun onDestroyView() {
         super.onDestroyView()
-        adUnit?.stopAutoRefresh()
```
YuriyVelichkoPI, line 104: This change looks as potential memory leak. What if the publisher utilizes a new counting methodology and will not call `destroy()`?

YuriyVelichkoPI (changes requested):

Example/PrebidInternalTestApp/src/main/java/org/prebid/mobile/renderingtestapp/plugplay/bidding/gam/original/GamOriginalInstreamNewApiFragment.kt
ValentinPostindustria: If a user uses this burl tracker, it's better to use destroy(). Otherwise the monitor will traverse view hierarchy as long as the ad view container exists.

YuriyVelichkoPI (approved):
LGTM

YuriyVelichkoPI:
The PUC still fires the `burl`. So we will create a follow-up ticket to support new imptrackers schema once it is ready.  For now publishers can customize the PUC in order not to track BURL if they want to use SDK native tracking. 

The default behavior is not changed. The SDK tracking is opt-in.

PR #805, opened 2024-11-25
Fix OpenRTB field

ValentinPostindustria:
Add the new OpenRTB fields:
- Global level `TargetingParams.setGlobalOrtbConfig()`.
- Imp level `adUnit.setImpOrtbConfig()`. 

Fix merging of arrays. Refactor OpenRTB merger. Add unit tests.


PrebidMobile/PrebidMobile-core/src/test/java/org/prebid/mobile/OpenRtbMergerTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+        JSONObject mergedJson = merge(request, openRtb);
+
+        assertJsonEquals(request, mergedJson.toString());
+    }
+
+    @Test
+    public void merge_differentTypes() throws JSONException {
+        String request = "{}";
+        String openRtb = fromResources("merge_all_types.json");
```
YuriyVelichkoPI, line 58: Please don't use files. It's much more convenient to see the JSONs right in the tests.

YuriyVelichkoPI (changes requested):

YuriyVelichkoPI (changes requested):
Please add an example of how `setGlobalOrtbConfig` is used in the demo apps.

Let's set the following fields via arbitrary OpenRTB Config:

- ext.myext.test = 1
- displaymanager = "Google"
- displaymanagerver = $CurrentGMASDKVersion

Please add an example of how `setImpOrtbConfig` is used in the demo apps.

Let's set the following fields via arbitrary OpenRTB Config:
- imp.bidfloor=10
- imp.banner.battr = [1,2,3,4]

YuriyVelichkoPI (approved):
LGTM

PR #808, opened 2024-12-12
Fix smoke tests 

ValentinPostindustria:
After implementing the rewarded ad unit functionality, our test app's rewarded examples wait for the close button to trigger the default timeout, which is 120 seconds. To complete smoke tests faster, we replaced the rewarded config IDs.


YuriyVelichkoPI (approved):
LGTM

PR #817, opened 2025-01-06
Do not use device size if there is no ad size

ValentinPostindustria:
The current implementation sends a device size as the desired size when no other ad sizes are provided. We removed the device size sending functionality.  
[Issue](https://github.com/prebid/prebid-mobile-ios/issues/391).


YuriyVelichkoPI (approved):
LGTM

PR #829, opened 2025-02-04
Auction settings id

mdanylov-sigma:
Issue [#946](https://github.com/prebid/prebid-mobile-ios/issues/946)

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
@@ -164,6 +164,11 @@ public abstract class AdUnit {
                 return;
             }
         }
+        if (PrebidMobile.getAuctionSettingsId() != null && TextUtils.isEmpty(PrebidMobile.getAuctionSettingsId())) {
+            LogUtil.error("Empty auction settings id.");
+            listener.onComplete(ResultCode.INVALID_AUCTION_SETTINGS_ID);
+            return;
+        }
 
         HashSet<AdSize> sizes = configuration.getSizes();
         for (AdSize size : sizes) {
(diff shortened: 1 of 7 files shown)


PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
```diff
@@ -164,6 +164,11 @@ public void fetchDemand(Object adObject, @NonNull OnCompleteListener listener) {
                 return;
             }
         }
+        if (PrebidMobile.getAuctionSettingsId() != null && TextUtils.isEmpty(PrebidMobile.getAuctionSettingsId())) {
```
ValentinPostindustria, line 167: It would be better to combine this if statement with the check above. 
```
if (TextUtils.isEmpty(PrebidMobile.getPrebidServerAccountId())) {
```
It's okay with this feature if there is no server account id but there is an auction settings id.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/mediation/MediationBaseAdUnit.java
```diff
@@ -98,6 +99,11 @@ protected void fetchDemand(@NonNull OnFetchCompleteListener listener) {
             listener.onComplete(FetchDemandResult.INVALID_CONFIG_ID);
             return;
         }
+        if (PrebidMobile.getAuctionSettingsId() != null && TextUtils.isEmpty(PrebidMobile.getAuctionSettingsId())) {
```
ValentinPostindustria, line 102: The same here.

ValentinPostindustria (changes requested):

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
mdanylov-sigma: Removed this check to align the implementation with updated iOS one.

ValentinPostindustria (approved):

Issue #849, opened 2025-04-08
java.lang.RuntimeException: Can't create handler inside thread during BannerAdUnit creation


sarbyn:
**Describe the bug**
In or project, the Bid requests are done in a Dispatchers.IO coroutine scope.

So, in this scope, we build InterstitialAdUnit or BannerAdUnit instance, and then call the fetchDemand method in order to call the backend with the new adUnit. It works fine until version 2.5.0: now it is impossible to start the requuest, because the BannerAdUnit requires a timer, and does not work in our coroutine scopes.

```
java.lang.RuntimeException: Can't create handler inside thread Thread[DefaultDispatcher-worker-9,5,main] that has not called Looper.prepare()
	at android.os.Handler.<init>(Handler.java:228)
	at android.os.Handler.<init>(Handler.java:130)
	at android.os.CountDownTimer$1.<init>(CountDownTimer.java:129)
	at android.os.CountDownTimer.<init>(CountDownTimer.java:129)
	at org.prebid.mobile.VisibilityMonitor$VisibilityTimer.<init>(VisibilityMonitor.java:74)
	at org.prebid.mobile.VisibilityMonitor.<init>(VisibilityMonitor.java:26)
	at org.prebid.mobile.AdUnit.<init>(AdUnit.java:74)
	at org.prebid.mobile.AdUnit.<init>(AdUnit.java:86)
	at org.prebid.mobile.BannerBaseAdUnit.<init>(BannerBaseAdUnit.java:33)
	at org.prebid.mobile.BannerAdUnit.<init>(BannerAdUnit.java:43)
	at xxx.xxxx.xxxxx.xxxxx.PrebidHelper.getPrebidAdUnits$library_debug(PrebidHelper.kt:54)
```

It seems related to https://github.com/prebid/prebid-mobile-android/pull/804

**To Reproduce**
1. Create an instance of BannerAdUnit in a Dispatchers.IO coroutine scope

**Expected behavior**
It is possible to create an instance of BannerAdUnit

**Smartphone (please complete the following information):**
 - Device: Any Android
 - Version : only on 2.5.0

sarbyn:
It happens also on Prebid 3.0.0

mdanylov-sigma:
Issue has been fixed in 3.0.1

PR #856, opened 2025-04-22
Visibility timer creation

mdanylov-sigma:
#849 
- Moves timer creation on the main thread.

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/VisibilityMonitor.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/VisibilityMonitor.java
@@ -3,6 +3,8 @@ package org.prebid.mobile;
 import android.app.Application;
 import android.content.Context;
 import android.os.CountDownTimer;
+import android.os.Handler;
+import android.os.Looper;
 import android.view.View;
 import android.view.ViewGroup;
 import android.webkit.WebView;
@@ -23,7 +25,12 @@ public class VisibilityMonitor {
 
     private boolean stopAfterFirstFinding = false;
 
-    private final VisibilityTimer visibilityTimer = new VisibilityTimer();
+    private VisibilityTimer visibilityTimer;
+
+    public VisibilityMonitor() {
+        Handler mainHandler = new Handler(Looper.getMainLooper());
+        mainHandler.post(() -> visibilityTimer = new VisibilityTimer());
+    }
 
     @Nullable
     private VisibilityActivityListener activityListener;


sarbyn:
Any ETA for this fix in the main branch? Thanks!

mdanylov-sigma:
> Any ETA for this fix in the main branch? Thanks!

Prioritised this on the PMC call, waiting for a review and going to release a patch.

ChrisHuie:
@jerrycychen and @cyu813 added you as reviewers to this repo. Can update and add you to this pr once you accept the invite

cyu813 (approved):
@mdanylov-sigma approved, thanks

ChrisHuie (dismissed):

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/VisibilityMonitor.java
```diff
@@ -23,7 +25,12 @@ public class VisibilityMonitor {
 
     private boolean stopAfterFirstFinding = false;
 
-    private final VisibilityTimer visibilityTimer = new VisibilityTimer();
+    private VisibilityTimer visibilityTimer;
+
+    public VisibilityMonitor() {
+        Handler mainHandler = new Handler(Looper.getMainLooper());
+        mainHandler.post(() -> visibilityTimer = new VisibilityTimer());
```
YuriyVelichkoPI, line 32: Would it potentially lead to crashes in the usage of `visibilityTimer` in other methods if it isn't created until then? 

For example, is the following will be safe with the current changes  (pseudo code): 

``` 
monitor = new VisibilityMonitor()
monitor.trackView(view, burl, cacheID)
```

If it's not safe, we should adjust the current fix to prevent further crashes if something has changed in the SDK due to the usage of  VisibilityMonitor.

sarbyn: BTW with this fix the issue #849 is fixed on my project, tested with a local build of **849-timer-inizialitation** branch

YuriyVelichkoPI: Yes, there are no objections to the fix itself, but we should also think about possible consequences if something has been changed in the code eventually. So the code should be as safe and as reliable as possible.

mdanylov-sigma: Would `if (visibilityTimer != null)` conditions inside the `VisibilityMonitor` be enough for now to avoid potential crashes ? Or do you have another idea how to run `visibilityTimer` on the main thread to resolve the original issue ?

YuriyVelichkoPI: Let's add `if (visibilityTimer != null)` to unblock the release and think about more elegant change in follow up work.

danilovmaxim: Thanks!

ChrisHuie (approved):

PR #873, opened 2025-07-23
Add init status code when not ok

magjoh380:
Add the HTTP status code to error message when prebid init is not ok.


YuriyVelichkoPI (approved):
LGTM

PR #886, opened 2025-09-22
Ad position for multiformat ad unit

### LINKED ISSUE #883: Ad position (`pos`) for multiformat `PrebidAdUnit`
**Describe the bug**
Current implementation of the multi-format `PrebidAdUnit` doesn't have setters for ad position, not for video, not for banner. But for all other ad units we have the ad position setters.

ValentinPostindustria:
Add separate ad position setter for BannerParameters and VideoParameters in multiformat ad unit. 
Closes #883.


mdanylov-sigma:
The [iOS implementation](https://github.com/prebid/prebid-mobile-ios/blob/47a425c455f50825bf62f51923daa47dd651331a/PrebidMobile/Swift/AdUnits/MultiformatAdUnit/PrebidRequest.swift#L25) has `adPosition` property in the `PrebidRequest` for this purposes, do we need to have a similar implementation for both platforms ?

PR #887, opened 2025-09-23
Is video check as on iOS

### LINKED ISSUE #769: Android SDK should rely on `ext.prebid.type` on selectong the rendering engine
## Motivation

SDK should be aligned with Prebid ecosystem.

## Background

Android SDK determines the bid format by [checking the content of the adm](https://github.com/prebid/prebid-mobile-android/blob/027fde2aafb9be39e792169a3be4eaf6c05de883/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/models/CreativeModelMakerBids.java#L82) filed. If it contains VAST tag the ad will be treated as `video`, otherwise as `banner`.

However, looking at the PBS codebase, we can see that the logic of determining the bid format is the responsibility of each particular bidder. There are different approaches to determining the bid type, and some implementations are pretty complex. 

So, it is not right to rely on internal SDK logic when the bidder knows the actual value better.

## Objectives 

The `BidResponse.isVideo()` function should check the value of `ext.prebid.type` response field instead of analyzing the content of the ad.

ValentinPostindustria:
Closes #769.

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/bidding/data/bid/Bid.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/bidding/data/bid/Bid.java
@@ -18,8 +18,10 @@ package org.prebid.mobile.rendering.bidding.data.bid;
 
 
 import android.util.Base64;
+
 import androidx.annotation.NonNull;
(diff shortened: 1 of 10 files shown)


YuriyVelichkoPI (approved):
LGTM

mdanylov-sigma (approved):
LGTM

Issue #858, opened 2025-05-01
SDK does not fire Prebid impression event URL in DISPLAY interstitial


elq81hc:
**Describe the bug**
SDK does not fire Prebid impression event URL in the DISPLAY Interstitial

**To Reproduce**
1. Enable [Events feature](https://docs.google.com/document/d/1ry0X4C2EV-R0pMrm1IQk9BstxaT395UCl3KKqTGa5c8/edit#heading=h.7w5yevygp2gz) on Prebid Server
2. Run example "In-App Display Interstitial"
3. Check whether the Prebid impression event URL fires or not

**Expected behavior**
Prebid impression event URL (ext.prebid.events.imp) should be fired when the ad is shown

**Smartphone (please complete the following information):**
 - Device: Pixel 8 Pro Simulator
 - OS: Android
 - Version: API 34

**Additional context**
- PrebidSDK version: 3.0.0
- Issue does not occur on the "Medium Phone" simulator

elq81hc:
Hi @YuriyVelichkoPI, please take a look at this issue

elq81hc:
Hi @YuriyVelichkoPI , @ValentinPostindustria, please review if it is an issue

mdanylov-sigma:
Hey @elq81hc! Thanks for highlighting this, we're going to check the described scenario!

elq81hc:
Hi @YuriyVelichkoPI may I know your though on this issue?

YuriyVelichkoPI:
Hi @elq81hc ! Sorry for the long silence, have a crazy week. 

Yep, it definitely looks like a bug. From the @mdanylov-sigma comment, I supposed that the ticket was considered in the Mobile PMC and prioritized for the work. 

Do you know if this bug is actual for iOS?

elq81hc:
Hi @YuriyVelichkoPI! Thanks for looking at this issue.

IOS does not face this issue. Tested on device iOS 15.8 and emulator IOS 17.3

BTW, how do I register to join Mobile PMC?

YuriyVelichkoPI:
If your company is a Prebid.org member, ask Katie Morgart <kmorgart@prebid.org> to add you to Prebid Mobile PMC meetings.

YuriyVelichkoPI:
We recently faced this issue as well. I re-assigned this issue to @ValentinPostindustria since he knows the reason and will provide the fix soon.

PR #899, opened 2025-11-11
Fix impression event 

ValentinPostindustria:
Original issue #858.


mdanylov-sigma (dismissed):

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/models/CreativeVisibilityTracker.java
```diff
@@ -87,21 +90,24 @@ public CreativeVisibilityTracker(
...
         this(trackedView, visibilityTrackerOptionSet);
         this.proceedAfterImpTracking = proceedAfterImpTracking;
     }
 
     public CreativeVisibilityTracker(
         @NonNull
         final View trackedView,
-        final VisibilityTrackerOption visibilityTrackerOption) {
+        final VisibilityTrackerOption visibilityTrackerOption
+    ) {
```
YuriyVelichkoPI: Please roll back code style changes.

PR #906, opened 2025-12-04
fix: hb_cache_id_local is not added to targetingKeywords if adObject is null

shinwan2:
### Summary
`GamOriginalApiNativeInApp` will fail to show the ad on 1bbd1da0bab74e878b4df4c5d34360c082222579 because `nativeAdUnit.fetchDemand((bidInfo) -> { …})` is used and `adObject` is `null`.

#### Why `adObject = null` fails

`NativeAdUnit` will need to propagate the `String cacheId` from `CacheManager.save(response.getWinningBidJson())` downstream so `AdViewUtils.findNative` can later find it inside one of Google Ad Manager classes (`NativeCustomTemplateAd`, `NativeCustomFormatAd`, etc). This propagation does not happen when `fetchDemand`'s `adObject` is `null` 
- `Utils.saveCacheId` returns early and does nothing
- Google Ad Manager instance does not have `hb_cache_id_local`
- `AdViewUtils.findNative` calls `PrebidNativeAdListener.onPrebidNativeNotValid`.

#### Why `adObject = AdManagerAdRequest` succeeds

The `Util.saveCacheId` with an instance of `AdManagerAdRequest` as `adObject` will add `hb_cache_id_local` into its `customTargeting` https://github.com/prebid/prebid-mobile-android/blob/8e431b6c15fe647de47d1f78e3952b3215ca388a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/Util.java#L380

That targeting is propagated into `NativeCustomFormatAd` so `AdViewUtils.findNative` can successfully get the saved ad from `CacheManager.get(cacheId)`.

### Solution

Add the `hb_cache_id_local` into `bidInfo.targetingKeywords` if it's native to be similar with `AdManagerAdRequest` code path.

### DIFF
--- a/Example/PrebidDemoJava/src/main/java/org/prebid/mobile/javademo/activities/ads/gam/original/GamOriginalApiNativeInApp.java
+++ b/Example/PrebidDemoJava/src/main/java/org/prebid/mobile/javademo/activities/ads/gam/original/GamOriginalApiNativeInApp.java
@@ -36,7 +36,6 @@ import org.prebid.mobile.javademo.activities.BaseAdActivity;
 import org.prebid.mobile.javademo.utils.ImageUtils;
 
 import java.util.ArrayList;
-import java.util.Map;
 
 public class GamOriginalApiNativeInApp extends BaseAdActivity {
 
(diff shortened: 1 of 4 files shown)


mdanylov-sigma:
> > I didn't change any version to cause the unit test failure 🤔
> > ```
> > GoogleAdVersionTest > checkIfLastVersionUsed FAILED
> >     org.junit.ComparisonFailure: Google Ad SDK was updated to 24.8.0! Please test Prebid SDK with the new version, resolve compilation problems, rewrite deprecated code. After testing you must update PrebidMobile.TESTED_GOOGLE_SDK_VERSION and also update version in publishing XML files (scripts/Maven/*.xml.) expected:<24.[8].0> but was:<24.[7].0>
> > ```
> 
> No worries. It means that the PBS team should upgrade the supported version of the GAM SDK before merging the PR. We'll just ask you to update your branch once it's done.

@shinwan2 #907 is merged, could u rebase your branch on master to resolve the failed pipeline issue

mdanylov-sigma (approved):
LGTM!

Issue #939, opened 2026-04-21
Bug: NPE on bid response error GMA-NextGen Integration


gm-ci-mobile:
**Describe the bug**
When using the GMA next-gen SDK for Android in combination with the Prebid Android SDK 3.3.1.

If the SDK encounters any bid response error here:

```java
// AdUnit.java
@Override
  public void onError(AdException exception) {
  LogUtil.error(TAG, "Bid response error: " + exception);
  bidResponse = null;

  Util.apply(null, adObject);
  originalListener.onComplete(convertToResultCode(exception));
}
```

The Util `apply` method then passes `null` into the bids parameter for [handleNextGenSdkTargeting](https://github.com/prebid/prebid-mobile-android/blob/0abbd45590282d844fc3a660884919e205ca39da/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/Util.java#L419) causing an NullPointerException when invoking java.util.Set java.util.HashMap.entrySet().

Full stack trace:

```
java.lang.NullPointerException: Attempt to invoke virtual method 'java.util.Set java.util.HashMap.entrySet()' on a null object reference 
at org.prebid.mobile.Util.handleNextGenSdkTargeting(Util.java:420)
at org.prebid.mobile.Util.apply(Util.java:360)
at org.prebid.mobile.AdUnit$1.onError(AdUnit.java:307)
at org.prebid.mobile.rendering.bidding.loader.BidLoader.failedToLoadBid(BidLoader.java:209)
at org.prebid.mobile.rendering.bidding.loader.BidLoader.-$$Nest$mfailedToLoadBid(Unknown Source:0)
at org.prebid.mobile.rendering.bidding.loader.BidLoader$1.onErrorWithException(BidLoader.java:90)
at org.prebid.mobile.rendering.networking.BaseNetworkTask.onPostExecute(BaseNetworkTask.java:100)
at org.prebid.mobile.rendering.networking.BaseNetworkTask.onPostExecute(BaseNetworkTask.java:38)
at android.os.AsyncTask.finish(AsyncTask.java:771)
at android.os.AsyncTask.-$$Nest$mfinish(Unknown Source:0)
at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:788)
at android.os.Handler.dispatchMessage(Handler.java:107)
at android.os.Looper.loopOnce(Looper.java:232)
at android.os.Looper.loop(Looper.java:317)
[...]

PR #940, opened 2026-04-21
Checks nullable bids for next gen custom targeting

mdanylov-sigma:
#939

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/Util.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/Util.java
@@ -417,12 +417,11 @@ public class Util {
     }
 
     private static void handleNextGenSdkTargeting(HashMap<String, String> bids, Object builder) {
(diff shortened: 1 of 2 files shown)


postindustria-code (dismissed):
LGTM!

YuriyVelichkoPI:
@mdanylov-sigma please double check if the fix is still needed? The issue was found before `cfd753d` which introduces the current guard 

```
if(bids == null || bids.isEmpty()) {
    return;
}
```

If so - let's keep the current version. It's better to have "guards" then code inside if  { }

mdanylov-sigma:
> @mdanylov-sigma please double check if the fix is still needed? The issue was found before `cfd753d` which introduces the current guard
> 
> ```
> if(bids == null || bids.isEmpty()) {
>     return;
> }
> ```
> 
> If so - let's keep the current version. It's better to have "guards" then code inside if { }

@YuriyVelichkoPI 
I've updated code, just added an unit test to verify the implementation.

YuriyVelichkoPI (approved):
LGTM

PR #941, opened 2026-04-27
Add didInjectView callback to PrebidMobilePluginRenderer

MattMur:
### Summary
This PR addresses a few issues I ran into while working with the plugin renderer system. 

**1. getPreferredPluginRendererName() returns null by default**
When a bid response doesn't include a renderer key in its meta, `getPreferredPluginRendererName()` is returning null. This  leads to a live production bug where the `if` condition in [BannerView.displayPrebidView](https://github.com/prebid/prebid-mobile-android/blob/f1fd15afc10a3685a643de85e1b8052cdcc9c94f/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/rendering/BannerView.java#L489-L494) that can never be hit, since by default it will always evaluate to false. 

**Fix:** Set PrebidRenderer as the default for `getPreferredPluginRendererName()`

**2. Final layout responsibility should be given to chosen PluginRenderer**
Layout parameters for the injected ad view are being set directly inside BannerView. Since BannerView doesn't have visibility into how a given renderer intends to display its content, this feels like the wrong place for this logic. The renderer itself is best positioned to own those decisions — it knows the ad format, creative dimensions, and any renderer-specific requirements.

**Fix:** Add new callback to the `PrebidMobilePluginRenderer` to notify it once the ad has been injected so it can manage the final layout.

**3. MATCH_PARENT is the wrong default for fixed-size banner ads**
Due to the bug from issue#1, MATCH_PARENT is always being set on the injected view. This is a potential source of issues since ideally the `PrebidRenderer` wants the size to be controlled by logic in `PrebidWebView`, and the outer `DisplayView` should simply WRAP_CONTENT.  

**Fix:** Use the new `didInjectView` method on PrebidRenderer to set WRAP_CONTENT with center gravity.

### Testing

I've tested these fixes on the PrebidInternalTestApp using every "Banner" "In-app" variant, and each one looks and behaves as expected.

Parallel PR for iOS [PR 1261](https://github.com/prebid/prebid-mobile-ios/pull/1261)

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/rendering/BannerView.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/rendering/BannerView.java
@@ -19,8 +19,6 @@ package org.prebid.mobile.api.rendering;
 import android.content.Context;
 import android.content.res.TypedArray;
 import android.util.AttributeSet;
-import android.util.Pair;
-import android.view.Gravity;
 import android.view.View;
 import android.widget.FrameLayout;
 import androidx.annotation.NonNull;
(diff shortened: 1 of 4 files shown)


postindustria-code (changes requested):
Please, add unit tests covering updated behaviour. Everything else LGTM

YuriyVelichkoPI:
@MattMur will you be able to add tests? If you don't have time for it now, let me know and we create follow-up ticket and merge this one

Issue #745, opened 2024-04-11
GAM Rendering API key values issue


acampbellthescore:
We're attempting to migrate from GAM original to GAM rendering API but have run into an issue accommodating all elements of our current stack into the new integration and are wondering how to proceed.

Currently for display ads, we send requests to Prebid, Amazon TAM and Nimbus to conduct auctions before building the GAM request object. Each SDK returns a key value pair that gets dynamically inserted into GAM request, along with our own set of custom key values that we use for our direct sold efforts.

Under the Rendering API setup, we're noticing that calls to the TAM and Nimbus SDKs no longer occur (which we can look into with our internal team), but the other issue we've noticed is those custom key value pairs we initially would build into the GAM request object aren't present with the rendering API set up. Would these key values need to be passed into the Prebid SDK instead since the Rendering API removes AdManagerAdRequest where these would normally be added?

YuriyVelichkoPI:
Hi @acampbellthescore ! I can confirm that Rendering API doesn't support passing "external" targeting parameters to GAM. So, no TAM or Nimbus line items can be targeted.  

It can be supported.

However, can you please share the demo code of how exactly your solution is built? No IDs are needed; we just need to inspect how and when you pass, collect, and pass the keywords.

acampbellthescore:
Hey @YuriyVelichkoPI, sorry for the delay in reply here. We checked with our team and they aren't able to provide demo code, unfortunately, but perhaps I can try to describe what we're looking for to transition to rendering API.

SInce we're a sports app, the key values we're passing into GAM are usually contextual to what page the user is viewing in the app. For example, if the user is viewing an NBA page for information on a specific game, we will pass information like `league = nba` and `match_id = 12345` and these are targetable key values in GAM for potential direct sold buys. There could be an application to add these into Prebid as First Party Inventory (Context) Data targeting parameters if DSPs can match against them, but ideally, we would want to still have these show in a GAM request so any line items we use these key values for are considered as part of the GAM auction and whether GAM determines it needs to render the ad instead of the Prebid SDK.

If the Nimbus/TAM use case could also be looked at, that would be very helpful for us in what we're trying to do.

Let me know if you have any follow up questions.

PR #944, opened 2026-05-15
Custom targeting for GAM rendering API

mdanylov-sigma:
[#1072](https://github.com/prebid/prebid-mobile-ios/issues/1072)
https://github.com/prebid/prebid-mobile-android/issues/745

### DIFF
--- a/PrebidMobile/PrebidMobile-gamEventHandlers/src/main/java/org/prebid/mobile/eventhandlers/AdManagerRequestConfiguration.java
+++ b/PrebidMobile/PrebidMobile-gamEventHandlers/src/main/java/org/prebid/mobile/eventhandlers/AdManagerRequestConfiguration.java
@@ -1,0 +1,32 @@
+/*
+ *    Copyright 2018-2025 Prebid.org, Inc.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package org.prebid.mobile.eventhandlers;
+
+import androidx.annotation.NonNull;
+
+import com.google.android.gms.ads.admanager.AdManagerAdRequest;
+
+/**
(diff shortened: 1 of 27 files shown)


postindustria-code:
Should this PR be updated to also apply the change to the Google Next-Gen handlers?

YuriyVelichkoPI (changes requested):
Nice fix — this closes a real gap from #745 (publishers losing their own custom key-values under the GAM Rendering API since `AdManagerAdRequest`/`BaseAdRequestBuilder` is now built internally by the event handler). The core mechanism is correct and well tested: the publisher's `configure()` callback runs first, then Prebid's own `hb_*` targeting is applied on top, so Prebid's reserved keys always win on collision — verified by `buildAdManagerRequest_WithConfigurationAndBid_MergesTargetingWithBidPriority` (GamUtilsTest.java) and `configureAdRequest_WithConfigurationAndBid_AppliesConfigurationBeforePrebidTargeting` (UtilsTest.kt). Good work also extending this to the Next-Gen handlers per the earlier review comment.

A few last nitpicks before merge — none are correctness bugs, just small polish items that would make this a cleaner public API to ship:

1. **Accessor naming consistency**: `GamBannerEventHandler`/`GamInterstitialEventHandler`/`GamRewardedEventHandler` correctly name their accessors after the interface (`setAdManagerRequestConfiguration`/`getAdManagerRequestConfiguration` for `AdManagerRequestConfiguration`). The Next-Gen handlers (`NextGenBannerEventHandler`/`NextGenInterstitialEventHandler`/`NextGenRewardedEventHandler`) drop the prefix (`setAdRequestConfiguration`/`getAdRequestConfiguration` for `NextGenAdRequestConfiguration`). Since these are new public setters, worth aligning them now while the PR is still open — e.g. rename to `setNextGenAdRequestConfiguration`/`getNextGenAdRequestConfiguration` to match the interface name, same as the GAM side.

2. **Undocumented priority semantics**: Neither `AdManagerRequestConfiguration.configure()` nor `NextGenAdRequestConfiguration.configure()` mentions that Prebid's own `hb_*` targeting will silently overwrite anything the publisher sets under the same key. A publisher trying to override `hb_pb` via this hook will be confused when it doesn't take effect — worth a one-line doc comment on each interface.

3. **Minor test fragility**: `UtilsTest.configureAdRequest_WithConfigurationAndBid_AppliesConfigurationBeforePrebidTargeting` asserts `putCustomTargeting("hb_pb", ...)` is invoked before `putCustomTargeting("hb_cache_id", ...)`, relying on `HashMap` iteration order for those two keys. Stable in practice on the JVM, but not guaranteed — consider verifying the calls without asserting relative order between the two Prebid keys.

Happy to approve once these are addressed — thanks for closing out #745!

YuriyVelichkoPI (approved):
LGTM

PR #945, opened 2026-05-19
Expire ads respectively to the seatbid.bid.exp

mdanylov-sigma:
[#850](https://github.com/prebid/prebid-mobile-ios/issues/850)

- Implemented `bid.exp` parsing and expiration handling for Native and Rendering API flows.
- Added Native cache expiration coverage, registered-view suppression, null listener behavior, and single-callback guarantees.
- Rendering ads now schedule expiration from `BidResponse`.
- Expiration timers are cancelled on show, destroy, and reload.
- Banner/interstitial/rewarded ads now notify expiration delegates.
- Expired interstitial/rewarded ads cannot be displayed.
- Covered Prebid and GAM expiration paths with unit tests.
- Split reusable test helpers into shared helper classes.
- Aligned `RewardedAdUnit.destroy()` listener cleanup with interstitial behavior.

### DIFF
--- a/Example/PrebidInternalTestApp/src/main/java/org/prebid/mobile/renderingtestapp/AdFragment.kt
+++ b/Example/PrebidInternalTestApp/src/main/java/org/prebid/mobile/renderingtestapp/AdFragment.kt
@@ -102,6 +102,7 @@ abstract class AdFragment : BaseFragment() {
         baseBinding?.root?.findViewById<EventCounterView>(R.id.btnAdClicked)?.isEnabled = false
         baseBinding?.root?.findViewById<EventCounterView>(R.id.btnAdDisplayed)?.isEnabled = false
         baseBinding?.root?.findViewById<EventCounterView>(R.id.btnAdClosed)?.isEnabled = false
+        baseBinding?.root?.findViewById<EventCounterView>(R.id.btnAdExpired)?.isEnabled = false
     }
 
     override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
--- a/Example/PrebidInternalTestApp/src/main/java/org/prebid/mobile/renderingtestapp/plugplay/bidding/base/BaseBidInterstitialFragment.kt
+++ b/Example/PrebidInternalTestApp/src/main/java/org/prebid/mobile/renderingtestapp/plugplay/bidding/base/BaseBidInterstitialFragment.kt
@@ -99,6 +99,14 @@ abstract class BaseBidInterstitialFragment : AdFragment(),
         binding.btnLoad.isEnabled = true
     }
 
+    override fun onAdExpired(interstitialAdUnit: InterstitialAdUnit?) {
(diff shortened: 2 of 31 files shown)


YuriyVelichkoPI (commented):
## Review

Implementation covers the mechanical parts of prebid-mobile-ios#850 well (bid.exp parsing, timer scheduling/cancellation across Banner/Interstitial/Rewarded, `isLoaded()`/`show()` gating, native cache expiration wiring, listener additions as non-breaking `default` methods, RewardedAdUnit.destroy() listener cleanup alignment) and has thorough test coverage (destroy-before-expiry, reload-cancels-old-timer, no-exp-means-no-timer, single-callback guarantee, null-listener safety, etc.).

A few things I'd like addressed before merging:

### 1. Expiration timer starts too late relative to bid.exp semantics
The issue spec says: *"Once the bid is received and the model is created... create an async call after the bid.exp time."* In `BaseInterstitialAdUnit`, `bidResponse` is assigned in `onFetchCompleted` (bid-receipt time), but `scheduleExpirationIfNeeded()` isn't invoked until `onInterstitialReadyForDisplay()` / `onAdServerWin()` — i.e. **after** the creative has finished downloading/rendering. Since `bid.exp` represents "seconds between auction and impression," anchoring the countdown to "ready to display" instead of "bid received" silently extends the effective window by the creative load time (potentially significant for VAST/video or heavy HTML creatives), undermining the billing-window enforcement this feature exists for. Please start the timer at bid-receipt time (or confirm against the iOS reference implementation which anchor point is intended, and document the choice).

### 2. Dead/unused plumbing in `CreativeModel` / `CreativeModelMakerBids`
`expirationTimeSeconds` is added to `CreativeModel` and populated in `parseAcj` (not for video/VAST models), but it is never read anywhere in production code — actual scheduling in `BannerView`/`BaseInterstitialAdUnit` reads `BidResponse.getExpirationTimeSeconds()` directly instead. The only consumers are round-trip getter/setter unit tests. Please either wire this up to something real or remove it to avoid confusing future maintainers.

### 3. Redundant condition in `BaseInterstitialAdUnit.show()`
`if (expired || !isAuctionWinnerReadyToDisplay())` — `isAuctionWinnerReadyToDisplay()` was itself updated to include `!expired`, so the `expired ||` is redundant. Please simplify to just `if (!isAuctionWinnerReadyToDisplay())`.

### 4. Banner "do nothing if not refreshable" still tears down the view
`expireAd()` in `BannerView` unconditionally calls `displayView.destroy()` + `removeAllViews()` regardless of `autoRefreshDelay`. The issue's "do nothing if ad unit is not refreshable" reads as "leave the currently displayed creative alone," but current behavior always blanks the view even when not refreshable. Please confirm this is the intended UX and adjust so the non-refreshable path truly does nothing to the displayed view (only fires the `onAdExpired` callback), or update the issue/PR description to clarify the deliberate deviation.

Requesting changes on all four points above — they all seem worth resolving before merge.

YuriyVelichkoPI (approved):
LGTM

Issue #872, opened 2025-07-23
device.geo.country is being sent in iso alpha-2 format


justadreamer:
**Describe the bug**
We noticed that device.geo.country is populated from carrier data and is sent in alpha-2 format, while OpenRTB spec demands it to be in alpha-3 format, see [here].(https://github.com/InteractiveAdvertisingBureau/openrtb2.x/blob/main/2.6.md#:~:text=ISO%2D3166%2D1-,%2Dalpha%2D3,-.)

**To Reproduce**
Intercept a request from a demo app running on a physical device using Charles and check device.geo.country. 
It is in alpha-2 format.

**Expected behavior**
device.geo.country should not be populated by the SDK from the carrier data as it is expected to be populated from reverse-geocoding the coordinates if they are sent on the server.  iOS does not populate this field.  However if we choose to populate it then it should be in alpha-3 format as OpenRTB spec requires.  

**Additional context**
Additionally @takenorim has pointed out the following details: 

I found the Prebid Mobile Android SDK uses Android's native API `getSimCountryIso` or `getNetworkCountryIso`. 
And, the API doc said it will return ISO-3166-1 alpha-2 country code (e.g. US)

In `prebid-mobile-android/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/GeoLocationParameterBuilder.java`: 
```
    private String getTelephonyCountry(Context ctx){
        TelephonyManager tm = (TelephonyManager) ctx.getSystemService(Context.TELEPHONY_SERVICE);


        if(tm != null) {
            String simCountry = tm.getSimCountryIso().toUpperCase();
            String networkCountry = tm.getNetworkCountryIso().toUpperCase();
```

The first one returns the ISO-3166-1 alpha-2 country code equivalent for the SIM provider's country code.
The second one returns the ISO-3166-1 alpha-2 country code equivalent of the MCC (Mobile Country Code) of the current registered operator or the cell nearby, if available. Note: Result may be unreliable on CDMA networks (use getPhoneType() to determine if on a CDMA network)

The following changes may fix it:

In `prebid-mobile-android/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/GeoLocationParameterBuilder.java`:
```
            String simCountry = tm.getSimCountryIso().toUpperCase();
            String networkCountry = tm.getNetworkCountryIso().toUpperCase();
```
=>

```
[...]

PR #947, opened 2026-05-19
Exclude device.geo.country from the request.

mdanylov-sigma:
#872 

## Summary

Fixes Android `device.geo.country` population to align with iOS behavior and OpenRTB expectations.

Previously, the SDK populated `device.geo.country` from carrier/locale data, which could produce ISO-3166-1 alpha-2 values such as `US` or `CA`. OpenRTB expects this field to use ISO-3166-1 alpha-3 when populated, and the value should generally come from location-based reverse geocoding rather than SIM/network carrier data.

## Changes

- Removed SDK-side population of `device.geo.country` from carrier, locale, and geocoder fallbacks.
- Added test coverage to verify `geo.country` is not populated from SIM/network carrier country values.
- Aligns Android behavior with iOS, where this field is not populated by the SDK.

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/GeoLocationParameterBuilder.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/GeoLocationParameterBuilder.java
@@ -16,23 +16,13 @@
 
 package org.prebid.mobile.rendering.networking.parameters;
 
-import android.content.Context;
-import android.location.Address;
-import android.location.Geocoder;
-import android.telephony.TelephonyManager;
-
-import org.prebid.mobile.LogUtil;
 import org.prebid.mobile.PrebidMobile;
 import org.prebid.mobile.TargetingParams;
 import org.prebid.mobile.Util;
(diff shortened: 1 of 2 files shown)


YuriyVelichkoPI (commented):
Why the population of country code was removed instead of converting it into the needed format using the system function?

mdanylov-sigma:
> Why the population of country code was removed instead of converting it into the needed format using the system function? 


Based on expected behaviour in the issue ->
image

Should we choose other way to resolve the issue? If yes, we need to align iOS to send
`geo.country` ?

YuriyVelichkoPI:
I've asked the issue owner about motivation to remove avaliable signal. Leet's wait for a feedback

postindustria-code (approved):
LGTM

YuriyVelichkoPI:
We discussed this ticket on PMC and agreed that we need more detailed investigation of what should be sent in the request. Different IAB standards require different format. 

Don't merge this one so far.

mdanylov-sigma:
Closing in favor of #1022. Following @YuriyVelichkoPI's decision in #872 ([comment](https://github.com/prebid/prebid-mobile-android/issues/872#issuecomment-5752350044)), we're keeping `device.geo.country` and converting it to alpha-3 behind an SDK-level config flag rather than removing it from the request.

PR #948, opened 2026-05-19
Fix banner load callbacks firing before creative attachment

mdanylov-sigma:
[#1220](https://github.com/prebid/prebid-mobile-ios/issues/1220)

## Summary

Fixed the banner/custom renderer load callback ordering on both iOS and Android so the load callback is not fired before the creative view is attached.
### Android

Updated `PrebidDisplayView.java` so `adLoaded` marks the load notification as pending instead of notifying immediately.

When `viewReadyForImmediateDisplay` is called, the creative is attached first, then pending `onAdLoaded` is fired, followed by `onAdDisplayed`.

Added regression coverage in `PrebidDisplayViewTest.java` to verify that:

- `onAdLoaded` is not fired before the creative is ready
- the creative is already attached when `onAdLoaded` fires
- `onAdLoaded` still fires before `onAdDisplayed`


postindustria-code (approved):
LGTM!

Issue #935, opened 2026-04-09
EIDs OpenRTB 2.6 support


pkowalski-id5:
## Type of issue
Intent to implement

## Description
Currently, the SDK places Extended Identifiers in `user.ext.eids`(deprecated [OpenRTB 2.5 convention](https://github.com/InteractiveAdvertisingBureau/openrtb/blob/main/extensions/2.x_official_extensions/eids.md)). 
In OpenRTB 2.6, `eids` was promoted to a first-class field on the `user` object (`user.eids`), 
and the EID object gained new provenance fields: `inserter`, `matcher`, `mm`. 
The SDK needs to support the OpenRTB 2.6 placement and provenance fields.

## Goals
- Add OpenRTB 2.6 provenance fields (`inserter`, `matcher`, `mm`) to `ExternalUserId`
- Add a typed `eids` field to `User.java` with proper serialization
- Write EIDs to `user.eids` (OpenRTB 2.6) by default
- Provide a opt-in flag to also write `user.ext.eids` (OpenRTB 2.5) for backwards compatibility

#### ExternalUserId - new OpenRTB 2.6 fields

New optional fields added to the existing model:

| Field | Type | Description |
|---|---|---|
| `inserter` | `String` | Domain of the entity that inserted this EID |
| `matcher` | `String` | Domain of the entity that performed the ID matching |
| `mm` | `Integer` | Match method (see AdCOM List: ID Match Methods) |

These fields are optional -- existing usage remains unchanged.

#### User model - new `eids` field

Add `eids` field directly to `User.java`, serialized in `getJsonObject()` as `user.eids`. Field type to be discussed: `JSONArray` (simple, consistent with the rest of the model) or `List<ExternalUserId>` (typed objects, better validation and readability -  requires serialization to JSON when building the request).

#### Configuration flag (PrebidMobile)

```java
@Deprecated
public static void setUseExtEidsBackwardsCompat(boolean enabled);

@Deprecated
public static boolean isUseExtEidsBackwardsCompat();
```

Default: `false` (only `user.eids`). When `true`, EIDs are written to both `user.eids` and `user.ext.eids`.

> **Open for discussion:** The default is `false` (opt-in for backwards compat). Publishers on OpenRTB 2.5 PBS must enable `setUseExtEidsBackwardsCompat(true)`. If too aggressive, the default can be flipped to `true`.

#### EID placement in BasicParameterBuilder

```
if eids exist:
[...]

PR #954, opened 2026-06-26
EID OpenRTB 2.6  support #935

pkowalski-id5:
[Closes](https://github.com/prebid/prebid-mobile-android/issues/935)


PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/BasicParameterBuilder.java
```diff
@@ -189,16 +190,27 @@ private void appendUserTargetingParameters(AdRequestInput adRequestInput) {
...
                     }
+                    Eid eid = id.toEid();
+                    if (eid != null) {
+                        eidList.add(eid);
+                    }
                 }
             }
-            user.getExt().put("eids", idsJson);
+            EidsPlacement placement = PrebidMobile.getEidsPlacement();
+            if (placement.inUserExt()) {
```
mdanylov-sigma, line 208: `user.getExt().put("eids", idsJson)` mutates the `Ext` instance returned by `TargetingParams.getUserExt()`, which is a shared static field reused across every ad request (not a fresh object per request). Once this line runs under `COMPATIBLE`/`OPEN_RTB_2_5`, the `"eids"` key is written into that shared instance permanently.

If a publisher later switches to `EidsPlacement.OPEN_RTB_2_6` to stop sending the 2.5 format, subsequent requests will still carry the stale `user.ext.eids` payload from the previous placement, because there's no corresponding `else { user.getExt().remove("eids"); }` (and same issue in reverse for `user.eids` if switching to `OPEN_RTB_2_5`).

PrebidMobile/PrebidMobile-core/src/test/java/org/prebid/mobile/rendering/networking/parameters/BasicParameterBuilderTest.java
```diff
@@ -125,6 +131,8 @@ public void cleanup() throws Exception {
         PrebidMobile.setPrebidServerAccountId("");
         PrebidMobile.setAuctionSettingsId(null);
 
+        PrebidMobile.setEidsPlacement(EidsPlacement.OPEN_RTB_2_6);
```
mdanylov-sigma, line 134: This resets `PrebidMobile.eidsPlacement` to `EidsPlacement.OPEN_RTB_2_6` instead of the documented default `EidsPlacement.COMPATIBLE` (see `PrebidMobile.java:142`). Since this is a static field and test JVMs are typically reused across test classes, this leaks non-default global state into whatever test runs next. Should be `PrebidMobile.setEidsPlacement(EidsPlacement.COMPATIBLE);` to match `setUp()` and the real default.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidMobile.java
```diff
@@ -137,6 +138,8 @@ public class PrebidMobile {
     @NonNull
     private static WeakReference<PrebidEventDelegate> eventDelegateReference = new WeakReference<>(null);
     private static boolean disableStatusCheck = false;
+    @NonNull
+    private static EidsPlacement eidsPlacement = EidsPlacement.COMPATIBLE;
```
mdanylov-sigma, line 142: Defaulting to `COMPATIBLE` means every existing integrator who upgrades without touching EID config now silently gets `user.eids` populated in addition to the `user.ext.eids`. 
The linked issue ([#935](https://github.com/prebid/prebid-mobile-android/issues/935)) proposed defaulting to `2.6-only`. 
Worth confirming this default flip is intentional and calling it out explicitly (changelog/migration notes), since it changes the payload for all current users with no opt-out short of an explicit `setEidsPlacement(OPEN_RTB_2_5)` call. 
Also, no test in `BasicParameterBuilderTest` currently exercises the non-default `OPEN_RTB_2_6/OPEN_RTB_2_5` paths against actual request output - only `COMPATIBLE` is verified.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/BasicParameterBuilder.java
```diff
@@ -189,16 +190,27 @@ private void appendUserTargetingParameters(AdRequestInput adRequestInput) {
             extendedIds.add(TargetingParams.getSharedId());
         }
         if (extendedIds != null && extendedIds.size() > 0) {
+            List<Eid> eidList = new ArrayList<>();
```
mdanylov-sigma, line 193: `id.getJson()` and `id.toEid()` (plus the `idsJson/eidList` containers) are built unconditionally for every `ExternalUserId`, before the `EidsPlacement` check.
In single-mode placements (`OPEN_RTB_2_6 `or `OPEN_RTB_2_5`), roughly half this per-ad-request work is computed and thrown away. Fix: check `placement.inUserExt()/placement.inUser()` before building each representation, not just before assigning it.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/models/openrtb/bidRequests/User.java
```diff
@@ -45,6 +50,14 @@ public JSONObject getJsonObject() throws JSONException {
         toJSON(jsonObject, "customdata", this.customData);
         toJSON(jsonObject, "geo", (geo != null) ? this.geo.getJsonObject() : null);
 
+        if (!CollectionUtils.isEmpty(eids)) {
```
mdanylov-sigma, line 53: `User.java`, `Eid.java` and `ExternalUserId.java` - these files import `com.google.android.gms.common.util.CollectionUtils` to check `!CollectionUtils.isEmpty(list)` on lists that are always non-null (new ArrayList<>(), never reassigned to null).

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/ExternalUserId.java
```diff
@@ -100,12 +142,53 @@ public JSONObject getJson() {
         }
     }
 
+    @Nullable
+    public Eid toEid() {
```
mdanylov-sigma, line 146: This duplicates the same serialization/validation rules in two separate paths: `getJson()` for `user.ext.eids` and `toEid()` for `user.eids`.

Both methods independently handle empty `source`, skip empty UID IDs, copy UID `atype`/`ext`, copy EID-level `ext`, return `null` when no valid UIDs remain, and now also map `inserter`/`matcher`/`mm`. That makes it easy for the two emitted representations to drift later if a field or validation rule is added to only one path.

Could we make one representation canonical, e.g. build a validated `Eid` once and serialize that for both `user.eids` and `user.ext.eids`? That would reduce migration risk, especially in compatible mode where publishers may compare the 2.5 and 2.6 payloads side by side.

mdanylov-sigma (changes requested):

mdanylov-sigma:
@pkowalski-id5 Thanks for your contribution! 
Are you planning to implement the same changes for iOS as well?

mdanylov-sigma:
Hey @pkowalski-id5 
Are you going to finalise this or should we take over ?

pkowalski-id5:
Hi @mdanylov-sigma, thanks for the comments, and sorry for the long delay. I had a holiday, and other priorities came up. I'm planning to get back to this early next week, but if this is urgent and can't wait, feel free to take it over.

mdanylov-sigma:
The feature was finalized in #1011 and has been merged for release in the next version. Thanks, @pkowalski-id5, for your efforts!

PR #955, opened 2026-06-29
Add optional PbScaleAppliedListener to AdViewUtils.findPrebidCreativeSize

Khushboo1028:
### Summary
Adds an **optional, additive** callback (`PbScaleAppliedListener`) to
`AdViewUtils.findPrebidCreativeSize(...)` that notifies the caller once the
corrective WebView initial-scale has been applied (the `fixZoomIn` step).

### Motivation
When resolving a Prebid creative's size, `findPrebidCreativeSize()` both reports
the size (`PbFindSizeListener.success`) and applies a corrective WebView scale via
`fixZoomIn` → `setWebViewScale(... setInitialScale ...)`. That correction runs
**after** the creative has already been laid out/painted at least once, and
`setInitialScale` only takes effect on the next reflow. As a result, the ad can
briefly render at the wrong (over-/under-zoomed) scale before snapping to the
corrected size — a visible flash.

Integrators currently have **no signal for when the corrective scale has been
applied**, so they cannot defer showing (e.g. fade in) the ad view until it will
paint at the right size. This PR provides that signal.

### Changes
- New `public interface PbScaleAppliedListener { void onScaleApplied(int width, int height); }`.
- New overload `findPrebidCreativeSize(View, PbFindSizeListener, PbScaleAppliedListener)`.
  The existing two-argument method is preserved and delegates with `null`.
- The optional listener is threaded through the internal async chain and invoked
  right after `setInitialScale(...)` (posted to the next frame, so it fires once
  the corrected scale will paint).
- The listener is only notified when a corrective scale is actually applied
  (gated on a positive content height); the existing scaling behavior is
  unchanged.

### Backward compatibility
Purely additive — **no breaking changes**:
- The existing `findPrebidCreativeSize(View, PbFindSizeListener)` signature and
  behavior are unchanged (it delegates with a `null` listener).
- Existing callers compile and behave exactly as before; the new behavior is
  reachable only by passing the new listener.
- Source- and binary-compatible (added methods/overload/interface only).

### Screenshot

Before

https://github.com/user-attachments/assets/c2a2bfff-012f-44ae-990b-0cac0ee08f64

After

https://github.com/user-attachments/assets/12e0b9ac-cf5b-4efa-9779-6f51bf6c2b85

### Usage
```java
AdViewUtils.findPrebidCreativeSize(
    adView,
    new AdViewUtils.PbFindSizeListener() {
        @Override public void success(int width, int height) { /* existing handling */ }
        @Override public void failure(@NonNull PbFindSizeError error) { /* ... */ }
    },
    (width, height) -> {
        // Corrective scale applied — safe to reveal the ad view now.
    });
onScaleApplied is best-effort: it is invoked at most once, and only when a
corrective scale is applied (not on the failure path or when no scale is needed).
Integrators that gate UI on it should keep a fallback (e.g. a timeout).
```

Testing

- Added unit tests in AdViewUtilsTest covering: the listener fires once with the
reported size on the success path; it is not notified for a non-positive content
height while the legacy setInitialScale call is preserved; and the legacy
(no-listener) path never posts a callback.
- Full PrebidMobile-core unit suite passes locally (1250 tests, 0 failures).

Before

Type of change

- [x] New feature / API addition (backward compatible)
- [ ] Bug fix
- [ ] Breaking change

Checklist

- [x] Backward compatible (existing API/behavior unchanged)
- [x] Unit tests added and passing
- [x] Tested on Yahoo Mail App

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/Util.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/Util.java
@@ -417,6 +417,9 @@ public class Util {
     }
 
     private static void handleNextGenSdkTargeting(HashMap<String, String> bids, Object builder) {
+        if(bids == null || bids.isEmpty()) {
+            return;
+        }
         Set<Map.Entry<String, String>> entries = bids.entrySet();
         for (Map.Entry<String, String> entry: entries) {
             Util.callMethodOnObject(builder, "putCustomTargeting", entry.getKey(), entry.getValue());
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/addendum/AdViewUtils.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/addendum/AdViewUtils.java
@@ -175,22 +175,22 @@ public final class AdViewUtils {
     }
 
     // Overload that applies the scale, then notifies the optional listener on the next frame (so it
-    // runs once the corrected scale has been laid out).
-    static void setWebViewScale(WebView webView, float webViewHeight, int webViewContentHeight, final int width, final int height, @Nullable final PbScaleAppliedListener scaleListener) {
-        //guard a non-positive content height: float/0 yields Infinity, and (int) Infinity is
-        //Integer.MAX_VALUE, which would set a nonsensical initial scale. Skip scaling (and the
-        //listener, since no corrective scale was applied) instead.
-        if (webViewContentHeight <= 0) {
-            return;
-        }
 
+    // runs once the corrected scale has been laid out). The scale computation and setInitialScale call
+    // are intentionally left identical to the legacy overload above so existing callers behave exactly
+    // as before; only the new, opt-in listener notification is added here.
+    static void setWebViewScale(WebView webView, float webViewHeight, int webViewContentHeight, final int width, final int height, @Nullable final PbScaleAppliedListener scaleListener) {
         //case: regulate scale because WebView.getSettings().setLoadWithOverviewMode() does not work
         int scale = (int) (webViewHeight / webViewContentHeight * 100 + 1);
 
         LogUtil.debug("Set WebView scale: " + scale + " (" + webViewHeight + ", " + webViewContentHeight + ")");
         webView.setInitialScale(scale);
 
-        if (scaleListener != null) {
+        // Notify the optional listener once the scale has been applied. Gate on a positive content
+        // height: a non-positive value makes the scale above nonsensical (float/0 -> Infinity ->
+        // Integer.MAX_VALUE), so we must not signal "scale applied" in that degenerate case. This only
+        // affects the new listener; the legacy (listener-less) path above is unchanged.
+        if (scaleListener != null && webViewContentHeight > 0) {
             webView.post(new Runnable() {
                 @Override
                 public void run() {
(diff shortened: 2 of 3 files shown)


Khushboo1028:
Re: the failing CI check (Run Unit Tests - Prebid Mobile)

The single failure is unrelated to this PR. It's GoogleAdVersionTest.checkIfLastVersionUsed in the PrebidMobile-gamEventHandlers module:

GoogleAdVersionTest > checkIfLastVersionUsed FAILED
  org.junit.ComparisonFailure: Google Ad SDK was updated to 25.4.0! ...
  expected:<25.[4].0> but was:<25.[2].0>

This is a canary test that fails whenever Google publishes a new GMA SDK version that's newer than Prebid's pinned PrebidMobile.TESTED_GOOGLE_SDK_VERSION. Google released 25.4.0 while upstream is still pinned to 25.2.0, so the check trips on master today and would fail on any open PR — it's not caused by this change. 

This PR's own tests pass: PrebidMobile-core:testReleaseUnitTest — the module containing the new AdViewUtils/PbScaleAppliedListener code and its unit tests — is green in the same CI run.

mdanylov-sigma:
> Re: the failing CI check (Run Unit Tests - Prebid Mobile)
> 
> The single failure is unrelated to this PR. It's GoogleAdVersionTest.checkIfLastVersionUsed in the PrebidMobile-gamEventHandlers module:
> 
> GoogleAdVersionTest > checkIfLastVersionUsed FAILED org.junit.ComparisonFailure: Google Ad SDK was updated to 25.4.0! ... expected:<25.[4].0> but was:<25.[2].0>
> 
> This is a canary test that fails whenever Google publishes a new GMA SDK version that's newer than Prebid's pinned PrebidMobile.TESTED_GOOGLE_SDK_VERSION. Google released 25.4.0 while upstream is still pinned to 25.2.0, so the check trips on master today and would fail on any open PR — it's not caused by this change.
> 
> This PR's own tests pass: PrebidMobile-core:testReleaseUnitTest — the module containing the new AdViewUtils/PbScaleAppliedListener code and its unit tests — is green in the same CI run.

@Khushboo1028 Could you please rebase the PR on the master branch to resolve the issue and pass all checks ?

Khushboo1028:
> > Re: the failing CI check (Run Unit Tests - Prebid Mobile)
> > The single failure is unrelated to this PR. It's GoogleAdVersionTest.checkIfLastVersionUsed in the PrebidMobile-gamEventHandlers module:
> > GoogleAdVersionTest > checkIfLastVersionUsed FAILED org.junit.ComparisonFailure: Google Ad SDK was updated to 25.4.0! ... expected:<25.[4].0> but was:<25.[2].0>
> > This is a canary test that fails whenever Google publishes a new GMA SDK version that's newer than Prebid's pinned PrebidMobile.TESTED_GOOGLE_SDK_VERSION. Google released 25.4.0 while upstream is still pinned to 25.2.0, so the check trips on master today and would fail on any open PR — it's not caused by this change.
> > This PR's own tests pass: PrebidMobile-core:testReleaseUnitTest — the module containing the new AdViewUtils/PbScaleAppliedListener code and its unit tests — is green in the same CI run.
> 
> @Khushboo1028 Could you please rebase the PR on the master branch to resolve the issue and pass all checks ?

Done, I think the workflow requires approval

mdanylov-sigma (approved):
LGTM!

mdanylov-sigma:
@Khushboo1028 just released v3.3.2 with your changes 🚀

Khushboo1028:
> @Khushboo1028 just released v3.3.2 with your changes 🚀

Thank you so much, really appreciate the promptness

Issue #950, opened 2026-05-21
AdBrowserActivity is exported="true"


andriiginting:
**Describe the bug**
`AdBrowserActivity` is declared with `android:exported="true"` in the SDK’s `AndroidManifest.xml`. This allows any third-party application on the device to launch the activity via an explicit `Intent`, potentially enabling arbitrary URLs or content to be displayed within the host app’s context.

**To Reproduce**

1. Install any app that integrates Prebid Mobile SDK
2. From a separate app on the same device, send an explicit Intent targeting `org.prebid.mobile.rendering.views.browser.AdBrowserActivity`
3. The activity launches within the host app's context

**Expected behavior**
`AdBrowserActivity` should only be launchable internally by the SDK itself (for example, when a user clicks an ad). It should not be accessible by external applications. The activity should therefore be declared with `android:exported="false"`.

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

mdanylov-sigma:
@andriiginting  just published v3.3.2

PR #964, opened 2026-07-07
Bid selector for winner override

mdanylov-sigma:
## Summary

- Adds `PrebidBidSelecting`, a functional interface letting publishers override the SDK's default marker-driven winning-bid selection, mirroring the equivalent iOS feature and resolving part 2 of [#950](https://github.com/prebid/prebid-mobile-ios/issues/950).
- The selector is evaluated **exactly once** per response, immediately after `seatbids` are parsed, and the resolved winner is cached on `BidResponse`:
 - The selector reference is snapshotted from `AdUnitConfiguration` at construction time, so mutating or clearing it afterward cannot change the winner of an already-returned response (e.g. one held by `BidResponseCache` or a slow-rendering flow).
 - `selectBid()` is invoked a single time — a stateful or non-deterministic selector could otherwise make `getWinningBid()`, `hasBidSelectorRejectedAllBids()`, `getTargeting()`, and `getWinningBidJson()` each observe a different "winner" for what should be one frozen auction result.
- Selector results are validated to be a member of the provided bids list.
- When a selector is active, non-winning bids' standard `hb_pb`/`hb_bidder`/`hb_cache_id` markers are stripped from the merged targeting; if the selector rejects all bids, targeting is empty and `AdUnit`, `NativeAdUnit`, `MultiformatAdUnitFacade`, and `MediationBaseAdUnit` all report `NO_BIDS` instead of `SUCCESS`.

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/AdUnit.java
@@ -277,6 +277,13 @@ public abstract class AdUnit {
             public void onFetchCompleted(BidResponse response) {
                 bidResponse = response;
 
+                if (response.hasBidSelectorRejectedAllBids()) {
+                    LogUtil.debug(TAG, "Bid selector rejected all bids.");
+                    Util.apply(null, adObject);
+                    originalListener.onComplete(ResultCode.NO_BIDS);
+                    return;
+                }
+
                 HashMap<String, String> keywords = response.getTargeting();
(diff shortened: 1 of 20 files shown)


mdanylov-sigma:
I put together a full proposal: problem statement, API shape (shown as pseudocode), design rationale for each non-obvious decision, alternatives considered, and open questions for maintainers. Posting it for review before this PR goes further.

[Full doc](https://docs.google.com/document/d/1CA6JjmiW0YKfM1qu3SRXjjPwESITzyYnNgnqWV2_52w/edit?usp=sharing), Part I.

PR #966, opened 2026-07-09
Gradle Cleanup Part 2 - Remove test.gradle build script

pablobaxter:
Removes the need for the `test.gradle` script and consolidates everything in the `libs.versions.toml` file instead.


mdanylov-sigma (approved):
LGTM

PR #976, opened 2026-07-29
Send ext.prebid.targeting flags as booleans, not strings

### LINKED ISSUE #961: Multiformat (Original API) sends `ext.prebid.targeting.includeformat` as string `"true"`, Prebid Server rejects auction with 400
**Describe the bug**
When building an **Original API** ad unit that contains **more than one ad format** (e.g. a multiformat banner + video interstitial), the SDK adds `ext.prebid.targeting.includeformat` to the bid request and serializes it as a **JSON string** (`"includeformat": "true"`) instead of a **JSON boolean** (`"includeformat": true`).

Prebid Server types `ext.prebid.targeting.includeformat` as a boolean and strict-decodes it (jsoniter), so the quoted string fails to unmarshal and **the entire auction is rejected with HTTP 400** — no bidders run:

```
Code 400. Invalid request: req.ext is invalid:
  cannot unmarshal ... ExtRequestTargeting.IncludeFormat: expect t or f
```

This is **version-independent** on Android — reproduced on 3.1.0, 3.3.1, and master. The same Prebid Server account works correctly from the **iOS SDK**, which sends the boolean form (`"includeformat": true`).

**Root cause (source)**

In the Original API request-building path, the targeting object is populated with a hardcoded **string** when the ad unit has multiple formats:

```java
// Prebid.java (targeting build), simplified:
if (isOriginalAdUnit() && getAdFormats().size() > 1) {
    Utils.addValue(targeting, "includeformat", "true");   // <-- string "true", not boolean true
}
```

`Utils.addValue(..., "true")` writes a `String`, so the serialized request carries `"includeformat":"true"`.

**To Reproduce**
1. Build an **Original API** ad unit (`AdUnit` / interstitial) configured with **two ad formats** (banner + video) so `getAdFormats().size() > 1`.
2. Run an auction against a Prebid Server that strict-decodes `ext.prebid.targeting.includeformat` as a boolean.
3. Observe the outgoing request `ext.prebid.targeting` contains `"includeformat":"true"` (quoted string).
4. Server responds **400 Invalid request** (`cannot unmarshal ... IncludeFormat: expect t or f`); the whole auction fails.

**Outgoing request (excerpt)**

```json
{
  "imp": [
    { "banner": { "...": "..." }, "video": { "...": "..." } }
  ],
  "ext": {
    "prebid": {
      "targeting": {
        "includewinners": true,
        "includebidderkeys": true,
        "includeformat": "true"
      }
    }
  }
}
```

`"includeformat": "true"` is a string; it should be the boolean `true`.

**Server response**

```
HTTP 400
Invalid request: req.ext is invalid:
  cannot unmarshal openrtb_ext.ExtRequestTargeting.IncludeFormat: expect t or f
```

**Expected behavior**

`ext.prebid.targeting.includeformat` should be serialized as a **JSON boolean** (`true`), matching:

- the OpenRTB / Prebid Server schema (it is a boolean), and
- the Prebid iOS SDK, which already emits the boolean and is accepted by the same server.

**Smartphone / environment**

- **Prebid Mobile Android SDK:** reproduced on 3.1.0, 3.3.1, and current master (version-independent)
- **API path:** Original API, multiformat ad unit (`getAdFormats().size() > 1`)
- **Server:** Prebid Server (Go) that strict-decodes `includeformat` as a boolean
- **Cross-check:** Prebid iOS SDK (same server account) sends boolean `true` → request accepted, auction runs

**Additional context**
- The field has no public setter and is added internally, so app code cannot change its type without an ORTB-level override.
- **Suggested fix:** emit a boolean in the multiformat branch, e.g. `Utils.addValue(targeting, "includeformat", true)` instead of the string literal `"true"` — matching the iOS SDK and the server's boolean type.

Madroid2:
Prebid Server declares the targeting flags as booleans in openrtb_ext.ExtRequestTargeting:

    IncludeWinners    *bool `json:"includewinners,omitempty"`
    IncludeBidderKeys *bool `json:"includebidderkeys,omitempty"`
    IncludeFormat     bool  `json:"includeformat,omitempty"`

The SDK was sending the JSON string "true" for all three, so unmarshalling fails server-side and the auction is rejected with HTTP 400.

includeformat is the visible case because it is set automatically whenever an original-API ad unit declares more than one format, so every multiformat request fails. includewinners and includebidderkeys carry the identical bug but default to false, so they only break for publishers who opt in via setIncludeWinnersFlag / setIncludeBidderKeysFlag.

Fixes #961

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/bidding/data/bid/Prebid.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/bidding/data/bid/Prebid.java
@@ -137,15 +137,15 @@ public class Prebid {
         JSONObject targeting = new JSONObject();
 
         if (config.isOriginalAdUnit() && config.getAdFormats().size() > 1) {
-            Utils.addValue(targeting, "includeformat", "true");
+            Utils.addValue(targeting, "includeformat", true);
         }
 
         if(PrebidMobile.getIncludeWinnersFlag()){
-            Utils.addValue(targeting, "includewinners", "true");
+            Utils.addValue(targeting, "includewinners", true);
         }
 
         if(PrebidMobile.getIncludeBidderKeysFlag()){
-            Utils.addValue(targeting, "includebidderkeys", "true");
+            Utils.addValue(targeting, "includebidderkeys", true);
         }
 
         Utils.addValue(prebid, "targeting", targeting);
--- a/PrebidMobile/PrebidMobile-core/src/test/java/org/prebid/mobile/rendering/bidding/data/bid/PrebidTest.java
+++ b/PrebidMobile/PrebidMobile-core/src/test/java/org/prebid/mobile/rendering/bidding/data/bid/PrebidTest.java
@@ -129,7 +129,7 @@ public class PrebidTest {
         config.addAdFormat(AdFormat.BANNER);
         config.addAdFormat(AdFormat.VAST);
         assertEquals(
-                "{\"storedrequest\":{\"id\":\"test\"},\"cache\":{\"bids\":{}},\"targeting\":{\"includeformat\":\"true\"}}",
+                "{\"storedrequest\":{\"id\":\"test\"},\"cache\":{\"bids\":{}},\"targeting\":{\"includeformat\":true}}",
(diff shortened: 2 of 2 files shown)


mdanylov-sigma (approved):
LGTM

PR #986, opened 2026-08-04
Fix controls misplaced in rewarded ads

### LINKED ISSUE #981: [Bug] Controls are misplaced in reward ad
**Describe the bug**
Progress indicator and mute button are misplaced. (see attached screenshot)
 
**To Reproduce**
1. Integrate Prebid Mobile Android 3.3.1 and load a rewarded video ad.
2. Render the bid response with `InterstitialController.loadAd(configuration, response)`.
3. Observe that the controls are placed correctly.

 **Expected behavior**
Progress indicator should be placed at left top corner.
mute button should be placed at right top corner. 
 
**Screenshots**

Image
 
 - Device: Samsung Galaxy S23
 - OS: Android 16
  - Version: Prebid Mobile Android 3.3.1

** Additional context:
this issue occurs also in iOS:
https://github.com/prebid/prebid-mobile-ios/issues/1298

mdanylov-sigma:
## Summary
Fixes controls being misplaced in rewarded video ads: the countdown/progress circle and the mute button both rendered near the bottom edge instead of their intended top corners, and could overlap when a skip/close button was configured to `TOP_LEFT`.

- Progress/countdown circle moved from bottom to top, matching the mute button.
- Mute and skip/close buttons switched from RTL-relative (`START`/`END`) to fixed (`LEFT`/`RIGHT`) gravity to stay consistent with the countdown circle's own fixed-corner layout, removing the corner mismatch that caused overlap.
- Added `moveCountdownPastTopLeftControl()` so the countdown circle shifts right when a `TOP_LEFT` skip/close button is configured, avoiding a new overlap in that combination.

Closes #981

### Screenshots
image-1785852518230

image-1785852521784

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/utils/helpers/Utils.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/utils/helpers/Utils.java
@@ -477,9 +477,9 @@ public final class Utils {
 
         FrameLayout.LayoutParams params;
         params = calculateButtonSize(view, buttonArea);
-        params.gravity = Gravity.END | Gravity.TOP;
+        params.gravity = Gravity.RIGHT | Gravity.TOP;
         if (buttonPosition == Position.TOP_LEFT) {
-            params.gravity = Gravity.START | Gravity.TOP;
+            params.gravity = Gravity.LEFT | Gravity.TOP;
         }
 
         view.setLayoutParams(params);
@@ -524,8 +524,7 @@ public final class Utils {
             ViewGroup.LayoutParams.WRAP_CONTENT,
             ViewGroup.LayoutParams.WRAP_CONTENT
         );
-        params.gravity = Gravity.END | Gravity.BOTTOM;
-        params.bottomMargin += 150;
+        params.gravity = Gravity.RIGHT | Gravity.TOP;
         view.setLayoutParams(params);
         InsetsUtils.addCutoutAndNavigationInsets(view);
         return view;
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/views/interstitial/InterstitialVideo.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/views/interstitial/InterstitialVideo.java
@@ -462,14 +462,54 @@ public class InterstitialVideo extends AdBaseDialog {
 
         RelativeLayout countdown = lytCountDownCircle.findViewById(R.id.lytCountdown);
         RelativeLayout.LayoutParams params = (RelativeLayout.LayoutParams) countdown.getLayoutParams();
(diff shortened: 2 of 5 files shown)


YuriyVelichkoPI (changes requested):
## Review: Fix controls misplaced in rewarded ads

**Diff summary:**
- `Utils.createButtonToNextPage()` (close/skip): `START/END` → `LEFT/RIGHT` gravity.
- `Utils.createSoundView()` (mute): moved from `END|BOTTOM` (+150px bottom margin) to `RIGHT|TOP`, margin removed.
- `lyt_countdown_circle_overlay.xml`: container gravity `bottom|start` → `top|left`; inner `lytCountdown` margin `bottom` → `top`.
- `InterstitialVideo.moveCountdownPastTopLeftControl()`: shifts the countdown circle right by the close/skip button's size + 10dp, but only when `buttonPosition == TOP_LEFT`.

The `START/END`→`LEFT/RIGHT` switch is justified: the countdown circle's XML margins were always hardcoded `Left`/non-RTL-aware, so keeping close/skip RTL-aware while the countdown wasn't caused the original mismatch — making all three fixed avoids that inconsistency.

**Concern — this PR introduces a new overlap it doesn't handle:**

In `AdBaseDialog`'s `OnDialogShowListener.onShow()` (unchanged by this PR), views are added to the same `FrameLayout` in this order: `addCloseView()` → `addSoundView()` → `addSkipView()`. In a `FrameLayout`, later-added siblings draw on top and win touch dispatch for overlapping regions.

- `lyt_close`/`lyt_skip` are 70dp, zero margin, gravity `RIGHT|TOP` (default position — no custom config needed).
- After this PR, `lyt_sound` is also zero-margin, gravity `RIGHT|TOP` — i.e., the same corner, by default.

Consequences once `isSoundButtonVisible=true` (opt-in, but a real usage) and the close/skip button is at its default `TOP_RIGHT` position (the common case, not `TOP_LEFT`):
- **No end card (`useSkipButton=false`):** `closeView` is added *before* `soundView`, so once the close button becomes visible (`InterstitialVideo.java:293`), it sits underneath the mute button in the same corner. Touches in that corner will hit the mute button, not the close button — the close button becomes effectively unusable.
- **With end card (`useSkipButton=true`):** `skipView` is added last so it stays clickable, but it fully visually covers the mute icon while shown.

`moveCountdownPastTopLeftControl()` only resolves the countdown-vs-close/skip collision for the `TOP_LEFT` custom-position case; it does nothing for this new sound-vs-close/skip collision in the default `TOP_RIGHT` case, which this diff itself introduces (the sound button didn't share a corner with anything before this PR).

**Test coverage:** New tests check `createSoundView` gravity in isolation and the existing close/skip gravity constant rename — none exercise the case where sound + close/skip are visible together, so this gap isn't caught.

**Requested change:** Please address the sound-vs-close/skip overlap in the default `TOP_RIGHT` configuration (e.g., offset sound away from close/skip similarly to how the countdown is offset, or ensure close/skip always stays topmost in z-order) before merging.

PR #989, opened 2026-08-04
Fix interstitial controls misplaced after orientation changes

### LINKED ISSUE #984: [Bug] Controls are misplaced when switch screen orientation in interstitial ad
**Describe the bug**
When screen orientation changes, controls are misplaced. See attached screenshot.
 
 **To Reproduce**

 1. Integrate Prebid Mobile Android 3.3.1 and load a interstitial ad.
  2. Render the bid response with InterstitialController.
  3. Show the interstitial ad.
  4. Change screen orientation(portrait -> landscape, vice versa)
  5. Observe that controls are displayed at appropriate positions.

**Expected behavior**
Controls should be placed at appropriate positions in any orientation.  
 
**Screenshots**

  
 - Device: Samsung Galaxy S23
 - OS: Android 16
 - Version: Prebid Mobile Android 3.3.1

 
**Additional context**
Disabling screen rotation in fullscreen ad might help mitigating the issue.

mdanylov-sigma:
## Summary
Fixes interstitial controls (close/skip/sound/countdown) being misplaced after device rotation, and fixes cumulative inset drift on repeated rotations.

- `InsetsUtils.applyInsetsToMargins` now applies the *delta* between the new and previously-applied insets, instead of resetting margins to a hardcoded value on every call — repeated rotations no longer cause controls to drift.
- Added RTL-aware gravity resolution (`Gravity.getAbsoluteGravity`) so `START`/`END`-configured views get insets applied to the correct physical side under RTL layout direction.
- `InterstitialView.refreshControlInsets()` (the renamed/extracted `onConfigurationChanged` logic) is now also invoked once at initial dialog show time, not only on rotation, closing a gap where controls were unadjusted until the first rotation.
- Bookkeeping fix: insets are only recorded as "applied" when margins were actually set on a supported `LayoutParams` type.

Closes #984

### Screenshots
image-1785852228998
image-1785852234273
image-1785852238144

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/rendering/InterstitialView.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/rendering/InterstitialView.java
@@ -54,15 +54,19 @@ public class InterstitialView extends BaseAdView {
     protected void onConfigurationChanged(Configuration newConfig) {
         super.onConfigurationChanged(newConfig);
 
+        refreshControlInsets();
+    }
+
+    public void refreshControlInsets() {
         List<View> views = Arrays.asList(
             findViewById(R.id.iv_close_interstitial),
             findViewById(R.id.iv_skip),
+            findViewById(R.id.iv_sound_interstitial),
             findViewById(R.id.rl_count_down),
             findViewById(R.id.tv_learn_more)
         );
 
         for (View view : views) {
-            InsetsUtils.resetMargins(view);
             InsetsUtils.addCutoutAndNavigationInsets(view);
         }
     }
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/rendering/InterstitialView.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/rendering/InterstitialView.java
@@ -275,7 +275,10 @@ public class InterstitialView extends BaseAdView implements ControlInsetsRefresh
         obstructionArray[4] = new InternalFriendlyObstruction(actionButton, InternalFriendlyObstruction.Purpose.OTHER, "Action button");
 
         if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP) {
-            View dialogRoot = closeInterstitial.getRootView();
+            // Use this view's own root rather than closeInterstitial's: closeInterstitial can be
+            // null if addCloseView() failed to add it (e.g. a cleared context reference), and
+            // this view reports the same root regardless.
+            View dialogRoot = getRootView();
             View navigationBar = dialogRoot.findViewById(android.R.id.navigationBarBackground);
             obstructionArray[5] = new InternalFriendlyObstruction(navigationBar, InternalFriendlyObstruction.Purpose.OTHER, "Bottom navigation bar");
         } else {
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/rendering/InterstitialView.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/rendering/InterstitialView.java
@@ -20,12 +20,15 @@ import android.content.Context;
 import android.content.res.Configuration;
 import android.util.Log;
 import android.view.View;
+import androidx.annotation.RestrictTo;
+import androidx.core.view.ViewCompat;
 import org.prebid.mobile.LogUtil;
 import org.prebid.mobile.api.exceptions.AdException;
 import org.prebid.mobile.configuration.AdUnitConfiguration;
 import org.prebid.mobile.core.R;
 import org.prebid.mobile.rendering.bidding.data.bid.BidResponse;
 import org.prebid.mobile.rendering.bidding.interfaces.InterstitialViewListener;
+import org.prebid.mobile.rendering.interstitial.ControlInsetsRefresher;
 import org.prebid.mobile.rendering.interstitial.DialogEventListener;
 import org.prebid.mobile.rendering.models.AdDetails;
 import org.prebid.mobile.rendering.models.internal.InternalFriendlyObstruction;
@@ -43,7 +46,7 @@ import java.util.List;
 /**
  * Internal view for {@link InterstitialAdUnit}.
  */
-public class InterstitialView extends BaseAdView {
+public class InterstitialView extends BaseAdView implements ControlInsetsRefresher {
 
     private static final String TAG = InterstitialView.class.getSimpleName();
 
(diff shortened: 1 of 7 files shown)


YuriyVelichkoPI:
The core insight is correct: `resetMargins()` hardcoded 16 **px** (not dp) and destroyed base margins from XML/code (e.g. `Utils.createSoundView`'s `bottomMargin += 150`), while `addCutoutAndNavigationInsets` was `+=`-accumulating. Removing the reset and making inset application idempotent is the right direction.

That said, the delta-based approach is fragile and there's one hard API-level bug.

## Blocking

**1. `isLayoutDirectionResolved()` is API 19, guarded at API 17 → crash on API 17–18**

`InsetsUtils.java:56` guards with `JELLY_BEAN_MR1` (17), but per the SDK's `api-versions.xml`:

```
<method name="isLayoutDirectionResolved()Z" since="19"/>
<method name="getLayoutDirection()I"        since="17"/>
```

`prebidSdkMinVersion = 16` (build.gradle:3), so this throws `NoSuchMethodError` on API 17/18. Lint isn't wired into CI (no `lintOptions` anywhere), so `NewApi` won't catch it. Simplest fix — drop the version dance entirely and use `ViewCompat.getLayoutDirection(view)` (`androidx.core.view` is already used in `Views.java`), or bump the guard to `KITKAT`.

**2. RTL fix is only half applied.** The `RelativeLayout` branch still maps `ALIGN_PARENT_END → rightMargin` and `ALIGN_PARENT_START → leftMargin`, which is inverted under RTL — the exact bug being fixed in the `FrameLayout` branch. `rl_count_down` goes through this branch.

## Design concerns

**3. Store base margins, not deltas.** The delta invariant "`previousInsets[side]` == what was applied to that side" only holds if the gravity gate is identical across calls. It isn't guaranteed: `previousInsets` is recorded as the *full* inset set even though only gravity-matching sides were applied, so any change in resolved gravity (RTL flip, `closeButtonPosition` change, `UNSPECIFIED_GRAVITY` params being replaced on re-`addView`) silently reintroduces drift — or worse, subtracts margins the code never added. Recording the *original* margins once and always computing `margin = base + inset` is fully idempotent and immune to all of that. It also removes the need for the `previousInsets` parameter.

**4. Static `APPLIED_INSETS` map.** Weak keys avoid the leak, but: `get()` + `put()` is a non-atomic read-modify-write, so `synchronizedMap` buys nothing real (and if it's main-thread-only, say so and drop the wrapper); and there's no way to reset it, so it's cross-test-pollutable global state. `view.setTag(R.id.prebid_base_margins, …)` scopes the state to the view's lifetime with no global state at all.

**5. `instanceof InterstitialView` in `AdBaseDialog` is a layering inversion** — `rendering.interstitial` now imports `api.rendering`. More importantly, is the `onShow` call actually needed? `Utils.createButtonToNextPage`/`createSoundView` already apply insets at creation using the **Activity's** decor view, which is attached and has valid insets at that point. Could you describe the concrete case where controls were unadjusted before the first rotation? If one exists, `InterstitialView` already receives `DialogEventListener.EventType.SHOWN` in `handleDialogEvent()` — reusing that hook would avoid the downcast.

**6. `refreshControlInsets()` is now public API** on a public class in `org.prebid.mobile.api.rendering`. Needs `@VisibleForTesting(otherwise = …)` or equivalent internal marking so it isn't treated as a supported entry point.

**7. `onConfigurationChanged` is not a reliable inset source.** Insets aren't guaranteed to be updated when the view-level config callback fires. Since the drift is now fixed, `setOnApplyWindowInsetsListener` on the container is the canonical API and would make this self-correcting; at minimum consider `view.post { … }`.

## Nits

- `InsetsUtils.resetMargins()` is now dead code — the only caller was the line removed here. Delete it.
- Magic `-1` → `FrameLayout.LayoutParams.UNSPECIFIED_GRAVITY`.
- `RelativeLayout` branch returns `true` on API < 23 even though the whole apply block is skipped, recording insets that were never applied.
- `iv_sound_interstitial` was added to `refreshControlInsets` but not to `formInterstitialObstructionsArray()` — intentional? The sound button is a real obstruction for OM.

## Tests

- **The new logic is untested.** Every `InsetsUtilsTest` case calls the `@VisibleForTesting` overload with hand-fed `previousInsets`, so the `APPLIED_INSETS` bookkeeping in `addCutoutAndNavigationInsets` — where the drift bug actually lived — has zero coverage. Consider extracting the inset *source* behind a seam so the public method can be tested with non-zero insets.
- `InterstitialViewTest.refreshControlInsets_UpdatesDialogSoundControlWithoutResettingMargins` is near-vacuous: Robolectric reports zero insets, so it only proves `resetMargins` isn't called. Rename to reflect that (`doesNotResetMarginsToDefault`) or make it meaningful.
- `applyInsetsWithEndGravityInRtl_UsesLeftInset` mocks `android.view.View` and then asserts on the very `params` object it stubbed `getLayoutParams()` with — the assertion can't fail for the wrong reason, but it's misleading. Prefer a real view with `@Config(qualifiers = "ldrtl")`. Also, once #1 is fixed, add an API 17/18 config case.
- Missing: `RelativeLayout` branch, `BOTTOM`/`LEFT` gravity, and a regression test that the sound view's 150px base bottom margin survives two rotations (the user-visible symptom).

YuriyVelichkoPI (changes requested):
see the comment above

YuriyVelichkoPI (changes requested):
## Blocking: `RelativeLayout.LayoutParams.getRule(int)` is API 23 — this reintroduces the round-1 crash class

Dropping the `Build.VERSION_CODES.M` guard in `InsetsUtils.applyInsetsToMargins()` isn't safe: that guard existed *because* of `getRule()`. Per the SDK's `api-versions.xml`:

```
<method name="getRule(I)I"  since="23"/>
<method name="getRules()[I"/>            <!-- API 1 -->
```

With `prebidSdkMinVersion = 16`, the RelativeLayout branch now throws `NoSuchMethodError` on API 16–22. Verified on this branch with a throwaway Robolectric test:

```
relativeLayoutBranch_onApi19 FAILED
    java.lang.NoSuchMethodError: 'int android.widget.RelativeLayout$LayoutParams.getRule(int)'
relativeLayoutBranch_onApi21 FAILED   (same)
```

This is not an edge case — it's on the path of **every VAST creative**:

`VideoCreative.showCallToAction()` → `VideoCreativeView.showCallToAction()` → `addCallToActionView()` → `InsetsUtils.addCutoutAndNavigationInsets(callToActionView)` → RelativeLayout branch

(`VideoCreativeView extends RelativeLayout`, and `tv_learn_more` hits the same branch again from `refreshControlInsets()`.) Since `NoSuchMethodError` is an `Error`, the surrounding `catch (Exception e)` blocks don't contain it — it propagates as a crash.

**Fix:** read the rules through `getRules()`, which is API 1 and backed by the same array:

```java
int[] rules = relativeParams.getRules();
boolean ruleTop    = rules[RelativeLayout.ALIGN_PARENT_TOP]    == RelativeLayout.TRUE;
boolean ruleBottom = rules[RelativeLayout.ALIGN_PARENT_BOTTOM] == RelativeLayout.TRUE;
boolean ruleRight  = rules[RelativeLayout.ALIGN_PARENT_RIGHT]  == RelativeLayout.TRUE;
boolean ruleLeft   = rules[RelativeLayout.ALIGN_PARENT_LEFT]   == RelativeLayout.TRUE;
boolean ruleEnd    = rules[RelativeLayout.ALIGN_PARENT_END]    == RelativeLayout.TRUE;
boolean ruleStart  = rules[RelativeLayout.ALIGN_PARENT_START]  == RelativeLayout.TRUE;
```

**Also needed:** `applyInsetsToMargins_OnLegacyApi_DoesNotCrash` cannot catch this — it only exercises the FrameLayout branch, and every RelativeLayout case runs under the class-level `@Config(sdk = 28)`. Please add a RelativeLayout case at `@Config(sdk = 19)`; it fails today and passes with the `getRules()` fix. (Minor: that test's comment says "API 17/18", but Robolectric can't run those — 19 is the floor.)

YuriyVelichkoPI (dismissed):
LGTM

YuriyVelichkoPI:
@mdanylov-sigma please solve the merge conflict

YuriyVelichkoPI (approved):

Issue #956, opened 2026-07-01
Deprecated API usage for device and user info


ValentinPostindustria:
The Prebid SDK collects device screen information and user geolocation; however, the current implementation relies on deprecated methods.

Issues identified:
* Location Services: `LastKnownLocationInfoManager` does not properly handle modern COARSE and FINE location permissions, relying instead on outdated permission models.
* Display Metrics: `DeviceInfoImpl` uses deprecated methods to retrieve screen width and height.
* Network Connectivity: `NetworkConnectionInfoManager` employs deprecated APIs to determine the network connection type.

PR #990, opened 2026-08-08
Read location providers according to the permission actually granted

Madroid2:
Change Log with reasoning:

- LastKnownLocationInfoManager gated both providers behind a single check that passed when either ACCESS_COARSE_LOCATION or ACCESS_FINE_LOCATION was granted, then read GPS_PROVIDER and NETWORK_PROVIDER inside one try block.

- GPS_PROVIDER requires precise location. Since Android 12 a user can grant approximate location on its own, and that is the common case -- the runtime dialog offers it as a first-class choice. In that state the GPS read throws SecurityException, which unwinds to the shared catch before NETWORK_PROVIDER is ever reached, so the SDK ends up with no location at all rather than the coarse fix the user did consent to. The only trace is a log line blaming an "android firmware bug", which sends anyone debugging it in the wrong direction.
 
- Each provider is now asked for separately and only when the permission it actually needs is held: GPS_PROVIDER for precise, NETWORK_PROVIDER for either grant. The read itself is isolated per provider, so a refusal or a provider missing on the device no longer prevents the other from being consulted -- IllegalArgumentException is handled for that second case, which the previous code did not cover at all. The selection logic that prefers GPS and falls back to the network fix via isBetterLocation is unchanged.
 
- Covers the new behaviour in LastKnownLocationInfoManagerTest: coarse-only never asks for GPS, a refused GPS read still yields the network location, and no grant reads no provider.
 
- Partially addresses #956 (location permissions). The deprecated display metrics and network connection APIs named in that issue are handled separately.

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/sdk/deviceData/managers/LastKnownLocationInfoManager.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/sdk/deviceData/managers/LastKnownLocationInfoManager.java
@@ -44,33 +44,54 @@ public final class LastKnownLocationInfoManager extends BaseManager implements L
         }
     }
 
-    @SuppressLint("MissingPermission")
     @Override
     public void resetLocation() {
-        Location gpsLastKnownLocation = null;
-        Location ntwLastKnownLocation = null;
-        if (getContext() != null) {
-            try {
-                locManager = (android.location.LocationManager) getContext().getSystemService(Context.LOCATION_SERVICE);
-
-                if (isLocationPermissionGranted() && locManager != null) {
-                    gpsLastKnownLocation = locManager.getLastKnownLocation(android.location.LocationManager.GPS_PROVIDER);
-                    ntwLastKnownLocation = locManager.getLastKnownLocation(android.location.LocationManager.NETWORK_PROVIDER);
-                }
-
-                if (gpsLastKnownLocation != null) {
-                    location = gpsLastKnownLocation;
-
-                    if (ntwLastKnownLocation != null && isBetterLocation(ntwLastKnownLocation, location)) {
-                        location = ntwLastKnownLocation;
-                    }
-                } else if (ntwLastKnownLocation != null) {
-                    location = ntwLastKnownLocation;
-                }
-            } catch (SecurityException exception) {
-                LogUtil.warning(TAG, "Unable to access locationManager due to android firmware bug.");
-            }
+        if (getContext() == null) {
+            return;
         }
+
+        locManager = (android.location.LocationManager) getContext().getSystemService(Context.LOCATION_SERVICE);
+        if (locManager == null) {
+            return;
+        }
+
+        // GPS_PROVIDER requires ACCESS_FINE_LOCATION; NETWORK_PROVIDER is satisfied by
+        // ACCESS_COARSE_LOCATION. Since Android 12 the user can grant approximate
+        // location on its own, so each provider is asked for separately and only when
+        // the permission it actually needs is held.
+        Location gpsLastKnownLocation = isFineLocationPermissionGranted()
+                ? lastKnownLocation(android.location.LocationManager.GPS_PROVIDER)
+                : null;
+        Location ntwLastKnownLocation = isLocationPermissionGranted()
+                ? lastKnownLocation(android.location.LocationManager.NETWORK_PROVIDER)
+                : null;
+
+        if (gpsLastKnownLocation != null) {
+            location = gpsLastKnownLocation;
+
+            if (ntwLastKnownLocation != null && isBetterLocation(ntwLastKnownLocation, location)) {
+                location = ntwLastKnownLocation;
+            }
+        } else if (ntwLastKnownLocation != null) {
+            location = ntwLastKnownLocation;
+        }
+    }
+
+    /**
+     * Reads a single provider in isolation. A provider that the platform refuses or
+     * that does not exist on this device must not stop the other provider from being
+     * read.
+     */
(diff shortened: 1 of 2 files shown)


YuriyVelichkoPI (commented):
Right bug, but the diagnosis is inverted and the fix overshoots. Details below.

### What's good

- The core defect is real: the two `getLastKnownLocation` calls shared one `try` block, so a `SecurityException` from GPS discarded the network read that never happened. Isolating each provider read in `lastKnownLocation(String)` fixes that properly.
- Removing the blanket `@SuppressLint("MissingPermission")` from `resetLocation()` and scoping it to the single risky call is the right move.
- Handling `IllegalArgumentException` (provider absent on the device) is legitimately new coverage.
- The early return on null context / null `locManager` reads better than the nested `if`.

### Main problem: the Android 12 claim is inverted

The PR description and the new code comment say the GPS read throws `SecurityException` when only approximate location is granted, "since Android 12". It is the other way around:

- **API < 31:** `LocationManagerService` enforced `checkResolutionLevelIsSufficientForProviderUse`, so coarse-only + `GPS_PROVIDER` threw `SecurityException`. This is where the bug actually bites.
- **API >= 31:** that check was replaced by coarsening through `LocationFudger`. A coarse-only client gets an obfuscated and throttled fix from *any* provider, GPS included, with no exception. The `LocationManager` javadoc was updated accordingly — the old "if your application only has the coarse permission then it will not have access to the GPS or passive location providers" sentence was dropped in favour of "the exact location will be obfuscated to a coarse level of accuracy".

Consequence: on Android 12+ with an approximate-only grant, the new `isFineLocationPermissionGranted()` gate skips a GPS fix that the platform *would* have returned, leaving only `NETWORK_PROVIDER`. Where the network fix is stale or the provider is absent, this converts a working (coarsened) location into no location at all — a regression on exactly the OS versions the PR claims to fix.

**Suggested change:** keep the per-provider `try` isolation, since that is the actual fix, but drop the fine-permission precondition. Gate both reads on `isLocationPermissionGranted()` and let the per-provider `catch` absorb the pre-31 refusal. Same defect fixed, nothing lost on 31+. If you prefer to keep the gate, condition it on `Build.VERSION.SDK_INT < Build.VERSION_CODES.S` and say so in the comment.

### Tests

- `Context` and `LocationManager` are both Mockito mocks, so the class-level `@Config(sdk = 19)` is inert here. The new tests assert the implementation's own branching rather than the platform premise: `verify(locationManager, never()).getLastKnownLocation(GPS_PROVIDER)` passes whether or not skipping GPS is correct on the target API. As written they lock in the behaviour that is questionable on 31+.
- `whenGpsProviderIsRefused_networkProviderIsStillRead` grants **fine** location and then makes the GPS read throw `SecurityException` — a state the framework does not produce. It does exercise the isolation, but the name and the scenario are misleading; the realistic case is coarse-only on API < 31.
- The test file still ends without a trailing newline.
- Each new test rebuilds its own `context` and manager instead of using the `@Before` fixture; a small `manager(coarseGranted, fineGranted)` helper would remove the triplication.

### Minor

- `isLocationPermissionGranted()` checks `getContext() != null` and then delegates to `isFineLocationPermissionGranted()`, which checks it again.
- `lastKnownLocation()` reads the `locManager` field instead of taking the manager as a parameter; passing it in would decouple the helper from field-assignment order.
- Pre-existing, but this PR is the natural place for it: `resetLocation()` never clears `location`, so once a fix is cached, a later permission revocation (runtime-revocable since Android 11) still leaves stale coordinates going out in bid requests via `GeoLocationParameterBuilder`. Consider setting `location = null` before the reads.

YuriyVelichkoPI (changes requested):

Madroid2:
@YuriyVelichkoPI You're right, and thank you for the detail , I had the platform behaviour backwards. I reasoned from the Android 12 approximate-location UX to "GPS must throw" without checking what the service actually does, and the LocationManager javadoc change you point to says the opposite: pre-31 checkResolutionLevelIsSufficientForProviderUse threw, and 31+ coarsens through LocationFudger instead. So the gate I added would have thrown away a usable coarsened fix on exactly the versions the PR claimed to help. b3d03cf.

I took your first option: dropped the fine-permission precondition entirely, gated both reads on isLocationPermissionGranted(), and let the per-provider catch absorb the pre-31 refusal. That keeps the part that was actually the fix  isolating each read so a SecurityException from GPS no longer discards a network read that never happened and costs nothing on 31+. I preferred it to the SDK_INT < S variant because the catch has to exist for the pre-31 path regardless, so the version check would be a second thing to keep correct for no behavioural gain. The comment now describes what the platform actually does, in both directions.

On the minor points: resetLocation() clears the cached fix before reading, so a revocation no longer leaves stale coordinates going out via GeoLocationParameterBuilder. lastKnownLocation() takes the manager as a parameter that left the locManager field with no readers, so I removed it. isLocationPermissionGranted() takes the context from its one caller instead of null-checking twice.

On the tests: you're right that the mocks make @Config(sdk = 19) inert, and I don't think I can fix that here with Context and LocationManager both mocked, these can only assert what the implementation does, not what the framework does. What I've changed is which behaviour they lock in. The verify(never()).getLastKnownLocation(GPS_PROVIDER) assertion is gone, since it was pinning the behaviour you flagged coarse-only now asserts both providers are asked. The refusal case is restated as coarse-only with GPS throwing, which is the state the framework actually produces, rather than fine-granted with a synthetic exception. Added a provider-absent case and one for the revocation behaviour above, folded the triplicated setup into a managerWith(locationManager, coarseGrant, fineGrant) helper, and added the trailing newline.

If you'd want the platform premise genuinely covered rather than documented, that would need an instrumented test on a real API 30 and API 31 device happy to open that separately if it's wanted, but it felt out of scope here.

PR #991, opened 2026-08-08
Bound click URL resolution and keep the click on network failure

### LINKED ISSUE #952: if the click link website is too slow, will not open the website page with web activity
**Describe the bug**

here is the click link, sdk will resolve this link and decide whether to redirect to web view activity or play video.

If a website loads slowly, it can cause the page to be unresponsive after a user clicks on it. 

Can you just open a web browser and load this URL? like ios sdk

https://apps-us2.adsgravity.io/v1/click?rid=ba51ffdf4fc225a5de5d79bdfb1d06aa91e9193621deaa4ac263386db0463741f9328f36611c39073276c99a290edc16229bc4f22bbdcf84e41ae321a86e0d61f9dc523a9a44aa8632ca7ebef75887900ef7cff0e86e9d8c5a6dd5482cca1444f52bc272f9065076a97ad74d8aaedf25ea5873601ff2b1ce72cf43249bc84c7a44d0d822d3707189b60ce1b14a1df5309fca4f0d7827cd0ed356ea8b639916adea1b800ed5577c8e7b4337330bc2bbd494a6bb545f43b706e2d10d8676aa67b83222c00ff7b1264cacbbd55b9d7b71d98e05ca74ee585e9d84677ccd34ce5db7574eb1a806b051be3d1403fa55d658560de95d9cacc3d9a9f010d28be9522101057c169c83566e4f900267de2a605e94a812cf9724c9a00b271faa8ed93260e1bf533737074ed38d158bdc64c77aaae9e2db504c37bdc2a3812c5f20f71d3a516405deb56555938cf3cf50a3f1b7b525641116ccd368e67b844e8b51bb36c6fc2aba0db0f433e8449d09a5303dd03296a45b6c41853a65f867221b3889618b92cad6557516a7a092d1f628c58ae6c67acdcb9d795516df1a4056853a740ba4aebe3949e2db777a549e24321be4577d5904bbe154cd79b780d613a49b79881aad42aa5bc33eb33b49b51852ca01c4f815786d7f24e4e7f1777fdf7e15a3963645885e020c95e0a0218a3c9bb91634b2078712c706f19606f8cc6f7b4af039216cae92826734ad6f3df845a1eb44c0c41eb9f4e73acbdece8767781bae457c0e65926b0d128e1006dfbb768bc08a2f328ee83c44ff3d1f4ac2b2794ec953c865485c55b2ce705eaf3a2c8ba1ad3d2042790e48310e78b09c02f05f0ad65385df970206e1bedb26d796805854bba3dc2fefff79caa3b3ab7185be2af79937a18c1c48c66c496e5a81d8296c268fab1d2a18734f4f62e4c3b93c53e578c65ee791db54a8d12312d239dfdd667c10589d19e860f69c60be6ce15534fb14ad842bff87a1fa1f82dfb95f9b2001e7df33293209dda4643186ca6830e3fa5685dac050fa71491a47570a1ac7b87a1f070e897955150c4d2c5dc9122259a7cbf08743133bec95117a43ccdeac15e5e2f6cef2d2aff0d06c31862316215de9263256787d77ba2083017e0a087ed840e77f496aa2fe394f28852ec00234b9d63a98e3fbe8986095cf80199069f4582bccdb8e4f4ee3cf7165d2c25dd6996c35fe187bb2bc930203dce3b8dd3efe9253727d9f5908aa44c295179890512b98afe497d61b41aa1334de64876f938bb998c0c25538289cd7e330a119798e4a2695cdd229ac330963b4885511b61a77dd43c43e06baaccff7cef5c09e76df76f6d7adadb8c9d81fa3a8e246d951748090bbfd22930ac114275b6814ded83e106a8d791c624ced84e8de7c554196ed29d71065d7e52b017f95a2a695622d268b4f2799cc1598fff2bb8e2995882b284ad54e8491424a7398e6b9a050611f9b7abdd758ce1f39750e2122c0eaa02567bb6fdace700eed2cdcf362f65459bcd81590ee0fde2d3be60a496fc00def0c002c782b88c75ac54430afb0a29025edd76259ab185f98b50e3e7d7bfc5eeb3634255f8ed6d54cedcdef318500e5f8111fb4063dd68c4345fc9259468d8d9eedffdce013b6b38332ad9f34a16c8995b3793db6e986f8a4713fd0b01350e41da4c4c2c51dddacaff8bfc5f9cdfaf983882ec07b79860d5b1e3aed8f956529965953c54fdc7889d1b1f3ba9fa1d7042f8a84e6ef1a539cdc979519a99ae3abf4517c4c2947f17cb5f00e1ed17029c1914451803c38a3f234702768d3c64d0d2f35ff23412a96189ca4def5a3a11f62eb9ff06db9ce7d1e18e028ae08b12af3d1cdc622aa218ba6d490cdd0a193ca81f28de63d14161b2e582f7481f596a034d38672d6a959ebe3645244fbf3f15bd885c4d7df628071039a3a82a222b5630e883e06148e3f86ba367feda7a6a694319c4b1003e115273f547c7c33da17e4e417aa38379b099575fbb039d45dedcfea2d2c162e7b8d92a16a0f5472c2d15a75f4998f

Madroid2:
Change Log with reasoning:

- UrlResolutionTask resolves a click-through by walking the redirect chain itself before handing a final URL to the browser. Two problems make a slow landing page swallow the click entirely.
 
- First, the connection is never bounded. HttpURLConnection defaults to no connect or read timeout at all, so a click-tracking host that accepts the socket and then stalls blocks doInBackground indefinitely. Nothing opens, and nothing reports a failure -- the tap simply does nothing.
 
- Second, an IOException anywhere in the chain discards the work already done: the catch returns null, onPostExecute treats null as a cancellation, and the click is dropped even though a perfectly usable URL was already in hand.
 
- This applies the timeouts BaseNetworkTask already uses for every other network call (PrebidMobile.getTimeoutMillis() for connect, the same value or SOCKET_TIMEOUT for read), and returns the last URL we had decided to follow when the chain fails mid-way. That URL is either the original click-through or a fully resolved hop, so handing it to the browser lets the browser finish the redirect chain instead of the user seeing nothing.
 
- The URISyntaxException path is deliberately left returning null. resolveRedirectLocation raises it when a redirect target is malformed, and the existing comment there is explicit that the request must stay cancelled rather than open an intermediary URL.

- Adds UrlResolutionTaskTest, which had no coverage: a normal redirect chain resolves to the final URL, a host that accepts and never answers now returns the click URL instead of null, deep links pass through untouched, and an empty argument list still returns null.
 
Fixes #952

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/mraid/methods/network/UrlResolutionTask.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/mraid/methods/network/UrlResolutionTask.java
@@ -22,6 +22,7 @@ import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
 import org.prebid.mobile.LogUtil;
 import org.prebid.mobile.PrebidMobile;
+import org.prebid.mobile.rendering.networking.BaseNetworkTask;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -65,9 +66,17 @@ public class UrlResolutionTask extends AsyncTask<String, Void, String> {
             }
         }
         catch (IOException e) {
-            return null;
+            // A slow or unreachable host in the middle of the redirect chain must not
+            // swallow the click. previousUrl is a location we already decided to follow
+            // -- either the original click-through or a fully resolved hop -- so handing
+            // it back lets the browser finish the chain instead of the tap doing nothing.
+            LogUtil.debug(TAG, "Url resolution stopped at " + previousUrl + ": " + e.getMessage());
+            return previousUrl;
         }
         catch (URISyntaxException e) {
+            // Deliberately not falling back here: resolveRedirectLocation throws this
+            // when the chain is malformed, and it must stay cancelled rather than open
+            // an intermediary URL.
             return null;
         }
         catch (NullPointerException e) {
@@ -86,6 +95,13 @@ public class UrlResolutionTask extends AsyncTask<String, Void, String> {
         try {
             httpUrlConnection = (HttpURLConnection) url.openConnection();
             httpUrlConnection.setInstanceFollowRedirects(false);
+            // HttpURLConnection defaults to no timeout at all, so a hanging redirect
+            // host blocks this task indefinitely and the click never opens anything.
+            // Bound it with the same values BaseNetworkTask already uses.
+            httpUrlConnection.setConnectTimeout(PrebidMobile.getTimeoutMillis());
+            httpUrlConnection.setReadTimeout(PrebidMobile.getTimeoutModified()
+                    ? PrebidMobile.getTimeoutMillis()
+                    : BaseNetworkTask.SOCKET_TIMEOUT);
 
             return resolveRedirectLocation(urlString, httpUrlConnection);
         }
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/mraid/methods/network/UrlResolutionTask.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/mraid/methods/network/UrlResolutionTask.java
@@ -26,12 +26,15 @@ import org.prebid.mobile.rendering.networking.BaseNetworkTask;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.ConnectException;
 import java.net.HttpURLConnection;
-import java.net.SocketTimeoutException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.net.UnknownHostException;
 import java.util.concurrent.TimeUnit;
+
+import javax.net.ssl.SSLHandshakeException;
 
 @VisibleForTesting
 public class UrlResolutionTask extends AsyncTask<String, Void, String> {
@@ -77,17 +80,27 @@ public class UrlResolutionTask extends AsyncTask<String, Void, String> {
                 redirectCount++;
             }
         }
-        catch (SocketTimeoutException e) {
-            // The reported case: the host accepted the socket and then stalled. The URL
-            // we were about to resolve is still worth opening, and the browser can
-            // finish the chain itself.
-            LogUtil.debug(TAG, "Url resolution timed out at " + previousUrl);
+        catch (UnknownHostException | ConnectException | SSLHandshakeException e) {
+            // The destination is known to be unreachable: DNS did not resolve, the
+            // connection was refused, or TLS could not be established. Opening the
+            // browser there would fail too, and it would record a click that never
+            // landed, so this stays cancelled.
+            LogUtil.debug(TAG, "Url resolution failed for " + previousUrl + ": " + e.getMessage());
+            return null;
+        }
+        catch (IOException e) {
+            // Everything else, including a stalled host and an abrupt disconnect. The
+            // JDK does not classify these consistently -- whether a reset surfaces as
+            // SocketException or as SocketTimeoutException depends on OS-level timing
+            // of the RST against the read -- so reachability must not be inferred from
+            // the exception type here. The URL in hand is one we had already decided to
+            // follow, so hand it to the browser and let it finish the chain.
+            LogUtil.debug(TAG, "Url resolution stopped at " + previousUrl + ": " + e.getMessage());
             return previousUrl;
         }
-        catch (IOException | URISyntaxException | NullPointerException e) {
-            // Hard failures -- unknown host, TLS, connection reset -- and malformed
-            // redirects stay cancelled. The destination is known to be unreachable or
-            // unsafe to open, so onFailure remains the correct outcome.
+        catch (URISyntaxException | NullPointerException e) {
+            // A malformed redirect target stays cancelled rather than opening an
+            // intermediary URL, as resolveRedirectLocation intends.
             return null;
         }
 
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/mraid/methods/network/UrlResolutionTask.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/mraid/methods/network/UrlResolutionTask.java
@@ -27,9 +27,11 @@ import org.prebid.mobile.rendering.networking.BaseNetworkTask;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.HttpURLConnection;
+import java.net.SocketTimeoutException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.util.concurrent.TimeUnit;
 
 @VisibleForTesting
 public class UrlResolutionTask extends AsyncTask<String, Void, String> {
@@ -49,6 +51,10 @@ public class UrlResolutionTask extends AsyncTask<String, Void, String> {
         }
 
         String previousUrl = null;
+        // One budget for the whole chain rather than per hop, so a slow host cannot
+        // multiply the delay by MAX_REDIRECTS before the browser is handed anything.
+        final long deadlineNanos = System.nanoTime()
+                + TimeUnit.MILLISECONDS.toNanos(resolutionBudgetMillis());
         try {
             String locationUrl = urls[0];
 
@@ -60,48 +66,60 @@ public class UrlResolutionTask extends AsyncTask<String, Void, String> {
                     return locationUrl;
                 }
 
+                long remainingMillis = TimeUnit.NANOSECONDS.toMillis(deadlineNanos - System.nanoTime());
+                if (remainingMillis <= 0) {
+                    LogUtil.debug(TAG, "Url resolution budget spent; opening " + locationUrl);
+                    return locationUrl;
+                }
+
                 previousUrl = locationUrl;
-                locationUrl = getRedirectLocation(locationUrl);
+                locationUrl = getRedirectLocation(locationUrl, (int) remainingMillis);
                 redirectCount++;
             }
         }
-        catch (IOException e) {
-            // A slow or unreachable host in the middle of the redirect chain must not
-            // swallow the click. previousUrl is a location we already decided to follow
-            // -- either the original click-through or a fully resolved hop -- so handing
-            // it back lets the browser finish the chain instead of the tap doing nothing.
-            LogUtil.debug(TAG, "Url resolution stopped at " + previousUrl + ": " + e.getMessage());
+        catch (SocketTimeoutException e) {
+            // The reported case: the host accepted the socket and then stalled. The URL
+            // we were about to resolve is still worth opening, and the browser can
+            // finish the chain itself.
+            LogUtil.debug(TAG, "Url resolution timed out at " + previousUrl);
             return previousUrl;
         }
-        catch (URISyntaxException e) {
-            // Deliberately not falling back here: resolveRedirectLocation throws this
-            // when the chain is malformed, and it must stay cancelled rather than open
-            // an intermediary URL.
-            return null;
-        }
-        catch (NullPointerException e) {
+        catch (IOException | URISyntaxException | NullPointerException e) {
+            // Hard failures -- unknown host, TLS, connection reset -- and malformed
+            // redirects stay cancelled. The destination is known to be unreachable or
+            // unsafe to open, so onFailure remains the correct outcome.
             return null;
         }
 
         return previousUrl;
     }
 
+    /** Read timeout for a single hop, mirroring {@link BaseNetworkTask}. */
+    private static int readTimeoutMillis() {
+        return PrebidMobile.getTimeoutModified()
+                ? PrebidMobile.getTimeoutMillis()
+                : BaseNetworkTask.SOCKET_TIMEOUT;
+    }
+
+    /** Total time allowed to resolve the chain: one connect plus one read. */
+    private static int resolutionBudgetMillis() {
+        return PrebidMobile.getTimeoutMillis() + readTimeoutMillis();
+    }
+
     @Nullable
-    private String getRedirectLocation(@NonNull final String urlString) throws IOException,
-                                                                               URISyntaxException {
+    private String getRedirectLocation(@NonNull final String urlString, final int budgetMillis)
+    throws IOException, URISyntaxException {
         final URL url = new URL(urlString);
 
         HttpURLConnection httpUrlConnection = null;
         try {
             httpUrlConnection = (HttpURLConnection) url.openConnection();
             httpUrlConnection.setInstanceFollowRedirects(false);
-            // HttpURLConnection defaults to no timeout at all, so a hanging redirect
-            // host blocks this task indefinitely and the click never opens anything.
-            // Bound it with the same values BaseNetworkTask already uses.
-            httpUrlConnection.setConnectTimeout(PrebidMobile.getTimeoutMillis());
-            httpUrlConnection.setReadTimeout(PrebidMobile.getTimeoutModified()
-                    ? PrebidMobile.getTimeoutMillis()
-                    : BaseNetworkTask.SOCKET_TIMEOUT);
+            // HttpURLConnection defaults to no timeout at all. Capping each hop at
+            // what is left of the budget also bounds the stream close below, which
+            // would otherwise wait a further read timeout after a stalled hop.
+            httpUrlConnection.setConnectTimeout(Math.min(PrebidMobile.getTimeoutMillis(), budgetMillis));
+            httpUrlConnection.setReadTimeout(Math.min(readTimeoutMillis(), budgetMillis));
 
             return resolveRedirectLocation(urlString, httpUrlConnection);
         }
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/mraid/methods/network/UrlResolutionTask.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/mraid/methods/network/UrlResolutionTask.java
@@ -25,16 +25,14 @@ import org.prebid.mobile.PrebidMobile;
 import org.prebid.mobile.rendering.networking.BaseNetworkTask;
 
 import java.io.IOException;
-import java.io.InputStream;
 import java.net.ConnectException;
 import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.UnknownHostException;
 import java.util.concurrent.TimeUnit;
-
-import javax.net.ssl.SSLHandshakeException;
 
 @VisibleForTesting
 public class UrlResolutionTask extends AsyncTask<String, Void, String> {
@@ -53,11 +51,22 @@ public class UrlResolutionTask extends AsyncTask<String, Void, String> {
             return null;
         }
 
+        // Read the timeout configuration once. PrebidMobile's statics are settable from
+        // any thread and this runs on THREAD_POOL_EXECUTOR, so re-reading them per hop
+        // against a deadline frozen here would let the two disagree mid-chain. Clamping
+        // at zero also keeps a negative value away from setConnectTimeout, which would
+        // throw.
+        final int configuredConnectMillis = Math.max(0, PrebidMobile.getTimeoutMillis());
+        final int configuredReadMillis = Math.max(0, BaseNetworkTask.readTimeoutMillis());
+        // Summed as a long: both are ints, and their sum overflows for a very large
+        // configured timeout, which would put the deadline in the past and stop the
+        // chain being resolved at all.
+        final long budgetMillis = Math.max(1L, (long) configuredConnectMillis + configuredReadMillis);
+
         String previousUrl = null;
         // One budget for the whole chain rather than per hop, so a slow host cannot
         // multiply the delay by MAX_REDIRECTS before the browser is handed anything.
-        final long deadlineNanos = System.nanoTime()
-                + TimeUnit.MILLISECONDS.toNanos(resolutionBudgetMillis());
+        final long deadlineNanos = System.nanoTime() + TimeUnit.MILLISECONDS.toNanos(budgetMillis);
         try {
             String locationUrl = urls[0];
 
@@ -76,25 +85,43 @@ public class UrlResolutionTask extends AsyncTask<String, Void, String> {
                 }
 
                 previousUrl = locationUrl;
-                locationUrl = getRedirectLocation(locationUrl, (int) remainingMillis);
+                locationUrl = getRedirectLocation(
+                        locationUrl, remainingMillis, configuredConnectMillis, configuredReadMillis);
                 redirectCount++;
             }
         }
-        catch (UnknownHostException | ConnectException | SSLHandshakeException e) {
-            // The destination is known to be unreachable: DNS did not resolve, the
-            // connection was refused, or TLS could not be established. Opening the
-            // browser there would fail too, and it would record a click that never
-            // landed, so this stays cancelled.
-            LogUtil.debug(TAG, "Url resolution failed for " + previousUrl + ": " + e.getMessage());
+        catch (MalformedURLException e) {
+            // A target java.net.URI accepts but java.net.URL cannot parse -- a
+            // non-numeric port, say. URI treats that authority as registry-based and
+            // resolves it happily, so the failure only surfaces on the next hop, and
+            // as an IOException rather than a URISyntaxException. It is a malformed
+            // target either way and must not reach an Intent.
+            LogUtil.error(TAG, "Malformed redirect target: " + previousUrl);
+            return null;
+        }
+        catch (UnknownHostException | ConnectException e) {
+            // DNS did not resolve, or the connection was refused. Neither depends on
+            // which HTTP stack asks, so the browser would fail here too; cancelling
+            // avoids recording a click that cannot land. This applies at any point in
+            // the chain: a hop confirmed unreachable is not worth handing over, even
+            // when an earlier hop succeeded.
+            //
+            // Deliberately not SSLHandshakeException. This probes with
+            // HttpURLConnection, which does not chase Authority Information Access for
+            // a server missing an intermediate certificate, while Chrome and WebView
+            // do -- so a handshake failure here does not predict one in the browser,
+            // and cancelling on it would drop clicks to pages a device renders fine.
+            LogUtil.debug(TAG, "Destination unreachable at " + previousUrl + ": " + e.getMessage());
             return null;
         }
         catch (IOException e) {
-            // Everything else, including a stalled host and an abrupt disconnect. The
-            // JDK does not classify these consistently -- whether a reset surfaces as
-            // SocketException or as SocketTimeoutException depends on OS-level timing
-            // of the RST against the read -- so reachability must not be inferred from
-            // the exception type here. The URL in hand is one we had already decided to
-            // follow, so hand it to the browser and let it finish the chain.
+            // Everything else: a stalled host, an abrupt disconnect, a TLS failure this
+            // stack is stricter about than a browser. The JDK does not classify these
+            // consistently -- whether a reset surfaces as SocketException or as
+            // SocketTimeoutException depends on OS-level timing of the RST against the
+            // read -- so reachability must not be inferred from the exception type. The
+            // URL in hand is one we had already decided to follow, so hand it to the
+            // browser and let it finish the chain.
             LogUtil.debug(TAG, "Url resolution stopped at " + previousUrl + ": " + e.getMessage());
             return previousUrl;
         }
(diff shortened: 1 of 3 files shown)


YuriyVelichkoPI (commented):
Sound direction overall — the timeout gap is real (`getRedirectLocation` was the only network call in the SDK with no timeouts) and the mirrored values match `BaseNetworkTask` lines 271-274 exactly. A few things worth resolving before merge.

### Needs discussion

**1. The `IOException` fallback silently converts a failed click into a *tracked* click.**

`previousUrl` flows to `onPostExecute` → `listener.onSuccess` → `UrlHandler.handleResolvedUrl` → `notifySuccess(url, trackingUrlList, urlAction)`. So click-tracking URLs now fire on the network-failure path, where previously `onFailure` fired. That's a measurable analytics change, not just a UX change, and it isn't mentioned in the description.

`IOException` is also broad — `UnknownHostException`, `SSLHandshakeException`, connection reset all now open the browser at a URL already known to be unreachable. Consider narrowing the fallback to `SocketTimeoutException` (the case in #952) and letting hard failures stay cancelled, consistent with the reasoning already applied to the `URISyntaxException` branch.

**2. This bounds the delay but doesn't eliminate it.**

Worst case is still per-hop rather than overall: `MAX_REDIRECTS = 3` × (2000 connect + 3000 read) ≈ 15s. The `finally` block's `getInputStream()` can also block for another `readTimeout` after a stalled hop, so in practice it can be worse. The reporter in #952 asked for the iOS behavior — hand the URL to the browser immediately. A single deadline checked across the whole `while` loop would match that intent much more closely than per-connection timeouts. At minimum it'd be good to note in the PR that the delay is bounded, not removed.

### Should fix

**3. The test leaks global static state.**

`PrebidMobile.setTimeoutMillis(300)` in `setUp` also flips `isTimeoutModified = true`, and nothing restores either value. Robolectric doesn't reset app statics between test classes sharing a sandbox, so subsequent tests' `BaseNetworkTask` read timeout becomes 300ms instead of `SOCKET_TIMEOUT`. There's no public reset — `WhiteBox` (already available via `:test-utils`) in an `@After` to restore `timeoutMillis = 2000` / `isTimeoutModified = false` would keep this contained.

### Minor

**4.** The headline case from the commit message — `IOException` on a *later* hop returning "a fully resolved hop" — isn't covered. Only first-hop failure is tested. A 302 → stalled-second-host case would assert the more interesting half of the change.

**5.** The `getTimeoutModified() ? getTimeoutMillis() : SOCKET_TIMEOUT` ternary is now duplicated in two places; extracting it would keep them from drifting. Worth noting this also repurposes `setTimeoutMillis`, documented as "connection timeout for bid request", for click resolution — reasonable precedent, but a publisher lowering it to shave auction latency now also shortens click resolution.

**6.** Comment volume is high relative to the surrounding style in this file; the "why" is already well captured in the commit message.

The test logic itself checks out — all four cases pass by inspection against the current `doInBackground`, `mockwebserver` is already in `libs.bundles.test.libraries`, and `@Config(sdk = 25)` / `@LooperMode(LEGACY)` match the sibling `GetOriginalUrlTaskTest`. I did not run the suite locally.

YuriyVelichkoPI (changes requested):

Madroid2:
@YuriyVelichkoPI  Both of the discussion points were fair. I've made changes as per the suggestions

1. IOException fallback. Narrowed to SocketTimeoutException. You're right that this was an analytics change and not just a UX one , I'd traced the path to the browser but not through notifySuccess to the tracking URLs. UnknownHostException, SSLHandshakeException and resets now stay cancelled and reach onFailure as before, which is also more consistent with the reasoning already applied to the URISyntaxException branch.

To be explicit about what remains: on the timeout path click tracking does now fire where it previously didn't. That's intended , the destination there is probably reachable and merely slow but it is a behaviour change, and it's now stated in the commit message and the PR description rather than left implicit.

2. Bounding. Replaced the per-hop timeouts with a single deadline across the whole loop: one connect plus one read for the entire resolution, with each hop capped at whatever remains. That also bounds the finally block's getInputStream(), which was the part that could make the worst case exceed the arithmetic. When the budget runs out the URL in hand goes to the browser instead of another hop being attempted. The delay is bounded, not removed , resolution is still attempted first , and the PR description now says that.

3. Test state leakage. tearDown restores timeoutMillis and isTimeoutModified through WhiteBox.

4. Added the mid-chain case , a 302 resolves before the second host stalls, so the URL returned is a resolved hop rather than the original. Also added a hard-failure case asserting the click stays cancelled, which pins the narrowing in point 1.

5. Extracted into readTimeoutMillis(), with resolutionBudgetMillis() alongside it.

On setTimeoutMillis being repurposed: that's a real consequence and the deadline makes it more so, since the overall budget now derives from it too. I kept it because reusing the existing knob seemed better than introducing a second one, and it matches BaseNetworkTask. If you'd rather click resolution had its own constant so publishers tuning auction latency don't shorten it, that's a small change , happy to make it.

6. Comment volume trimmed.

mdanylov-sigma:
The `Run Unit Tests - Prebid Mobile` CI job is failing here: https://github.com/prebid/prebid-mobile-android/actions/runs/32674172374/job/97481890562

```
UrlResolutionTaskTest > whenHostFailsHard_staysCancelled FAILED
    java.lang.AssertionError: expected null, but was:<http://localhost:55379/click>
```

**Root cause**

`whenHostFailsHard_staysCancelled` uses `SocketPolicy.DISCONNECT_AT_START` and asserts that `doInBackground()` returns `null`, on the assumption that an abrupt disconnect will surface as a plain `IOException`/`SocketException` (the "hard failure" branch in `UrlResolutionTask#doInBackground`), never as a `SocketTimeoutException` (the "stalled host" branch that falls back to the last resolved URL).

That assumption doesn't hold on the GitHub Actions runner: when the mock server closes the connection immediately after accept, `HttpURLConnection#getResponseCode()` there apparently blocks until the configured read timeout and throws `SocketTimeoutException` rather than failing fast with a connection-reset error. Since `SocketTimeoutException` is caught first in `doInBackground`, this hard-disconnect case gets routed into the "stalled host" fallback and the original click URL is returned instead of `null` — hence the assertion failure. It's timing/OS-dependent (I couldn't reproduce it every time locally), which is exactly why it's flaky rather than deterministically failing.

**Suggested changes**

1. Don't rely on "is it a `SocketTimeoutException`" as the signal for "hard vs. stalled" failure — the JDK doesn't guarantee that classification for every kind of abrupt disconnect (it depends on OS-level timing of FIN/RST vs. the read call). Two ways to make this robust:
   - Treat all `IOException`s the same way in `doInBackground` (fall back to `previousUrl` if one exists, else `null`), dropping the special-cased `SocketTimeoutException` catch, since the JDK can't reliably tell these apart; or
   - If the stalled-vs-hard-failure distinction is genuinely wanted, key it off exception types that *are* reliable signals of "destination unreachable" (`UnknownHostException`, `ConnectException`) rather than treating "not a `SocketTimeoutException`" as the hard-failure signal.

2. Regardless of which behavior you keep, `whenHostFailsHard_staysCancelled` shouldn't depend on real-socket timing to produce a guaranteed non-timeout exception. Consider either:
   - Testing `getRedirectLocation`/`resolveRedirectLocation` directly with a mocked `HttpURLConnection` that throws the specific exception type you want to assert on (no real sockets, no timing race), or
   - Using a scenario that deterministically produces a `ConnectException` (e.g. pointing at a closed/unbound local port) instead of `SocketPolicy.DISCONNECT_AT_START` on a live `MockWebServer`, which affects behavior only *after* connect and is where the race lives.

Happy to help test a revised version if useful.

Madroid2:
@mdanylov-sigma  Good catch, and thank you for the diagnosis , you're right that the fault was in the production code, not just the test. I was treating "not a SocketTimeoutException" as the hard-failure signal, which the JDK doesn't guarantee for abrupt disconnects. 1db5f3c.

I took your second suggestion rather than collapsing all IOExceptions, since @YuriyVelichkoPI  point stands that a click to an unreachable destination shouldn't open the browser or fire click tracking. Cancellation is now keyed off UnknownHostException, ConnectException and SSLHandshakeException, which are reliable "could not reach it at all" signals everything else, including stalls and ambiguous resets, falls back to the URL already in hand.

The test now points at a closed local port, which refuses the connection and gives ConnectException on every platform, instead of an abrupt disconnect from a live server where the race lives. Renamed to whenDestinationIsUnreachable_staysCancelled since that's the property being asserted. Thanks for the offer to help test , the run on this commit should tell us.

mdanylov-sigma (commented):
Automated review pass (Claude Code) on this diff. 11 findings below, most-severe first — 5 confirmed correctness bugs, plus a few lower-confidence/plausible items and cleanup notes. Two are empirically verified with standalone JVM tests (noted inline).

Overall: the timeout-and-fallback mechanism is a real improvement over "no timeout at all," but a few of the new exception-handling branches and the shared-budget math don't quite deliver on what their own comments claim, and two are demonstrably wrong (verified with a runnable test). Details on each line below.

**One more, not anchorable inline since `UrlHandler.java` isn't part of this diff — [worth confirming, likely intentional]:** `UrlHandler.java:225` (`TrackingManager.getInstance().fireEventTrackingURLs(trackingUrlList);`). Before this PR, a stalled/failed hop made `UrlResolutionTask` return `null`, which `onPostExecute` treated as cancellation — `onFailure` fired, no tracking. After this PR, the same stall returns `previousUrl` (non-null) → treated as success → `handleResolvedUrl` → `notifySuccess` → this line fires click/billing tracking pixels for a destination that was never confirmed to load. Very likely the intended consequence of "keep the click" (tracking-on-handoff, not tracking-on-confirmed-page-load, is how this codebase already behaves for successful resolutions) rather than an oversight — but it's a real, user-visible billing behavior change that wasn't called out in the PR description, so flagging for explicit confirmation.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/mraid/methods/network/UrlResolutionTask.java
```diff
@@ -59,33 +69,70 @@ protected String doInBackground(@Nullable String... urls) {
...
+    private static int resolutionBudgetMillis() {
+        return PrebidMobile.getTimeoutMillis() + readTimeoutMillis();
+    }
+
     @Nullable
-    private String getRedirectLocation(@NonNull final String urlString) throws IOException,
-                                                                               URISyntaxException {
+    private String getRedirectLocation(@NonNull final String urlString, final int budgetMillis)
+    throws IOException, URISyntaxException {
         final URL url = new URL(urlString);
```
mdanylov-sigma, line 125: **[CONFIRMED — empirically verified]** A redirect target that `java.net.URI` accepts but `java.net.URL` rejects will throw `MalformedURLException` here on the *next* hop. That's an `IOException` subtype, not `URISyntaxException`, so it lands in the `catch (IOException e)` branch (line 91) and returns `previousUrl` — not the `catch (URISyntaxException | NullPointerException e)` branch (line 101) whose comment says malformed targets must stay cancelled.

Verified with a standalone JVM test:
```java
new URI("http://example.com/a").resolve("http://example.com:abc/landing")
// -> succeeds, returns "http://example.com:abc/landing"
new URL("http://example.com:abc/landing")
// -> throws MalformedURLException: Error at index 0 in: "abc"
```
So a server that sends `Location: http://example.com:abc/landing` (non-numeric port — a real-world misconfiguration, not just a contrived example) causes the malformed string itself to be returned and handed to the browser Intent, contradicting the intent stated in the comment at line 102-103.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/mraid/methods/network/UrlResolutionTask.java
```diff
@@ -59,33 +69,70 @@ protected String doInBackground(@Nullable String... urls) {
...
+                }
+
                 previousUrl = locationUrl;
-                locationUrl = getRedirectLocation(locationUrl);
+                locationUrl = getRedirectLocation(locationUrl, (int) remainingMillis);
                 redirectCount++;
             }
         }
-        catch (IOException e) {
+        catch (UnknownHostException | ConnectException | SSLHandshakeException e) {
```
mdanylov-sigma, line 83: **[CONFIRMED]** This catch always returns `null`, regardless of how many redirects already succeeded — but `previousUrl` here holds the exact same kind of value ("a URL we already decided to follow") that the sibling `catch (IOException e)` below returns via `previousUrl` for e.g. a `SocketException` at the identical point in the loop. So the same variable state is treated inconsistently purely based on which `IOException` subtype was thrown.

Concretely: hop 1 (the tapped URL) succeeds and 302s to hop 2; fetching hop 2 throws `ConnectException`. This fully cancels the click (`onFailure` fires, nothing opens) rather than handing hop 2 to the browser the way a `SocketException` at the same point would. The new test suite only exercises this catch with `redirectCount==0` (`whenDestinationIsUnreachable_staysCancelled`), never after a prior hop already succeeded, so this asymmetry ships untested.

**Related, same block [CONFIRMED]:** `SSLHandshakeException` is treated as "destination is known to be unreachable... opening the browser there would fail too" — but this probes with plain `HttpURLConnection` (stock Android JSSE/Conscrypt), which unlike Chrome/WebView's TLS stack does **not** do AIA (Authority Information Access) chasing for a server missing an intermediate CA cert. That's a well-documented, common server misconfiguration (see OkHttp's own docs on why they ship `CertificateChainCleaner`) that a real mobile browser tolerates but this probe won't — so a landing page an actual device would render fine can get its click permanently dropped here.

**Also worth a look:** `NoRouteToHostException` and `SSLPeerUnverifiedException` are at least as "known unreachable" per this comment's own rationale, but aren't in this catch list — they fall through to the generic branch below instead. Lower-impact than the above since it errs toward "try anyway," but inconsistent with the stated logic.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/mraid/methods/network/UrlResolutionTask.java
```diff
@@ -59,33 +69,70 @@ protected String doInBackground(@Nullable String... urls) {
...
         final URL url = new URL(urlString);
 
         HttpURLConnection httpUrlConnection = null;
         try {
             httpUrlConnection = (HttpURLConnection) url.openConnection();
             httpUrlConnection.setInstanceFollowRedirects(false);
+            // HttpURLConnection defaults to no timeout at all. Capping each hop at
+            // what is left of the budget also bounds the stream close below, which
+            // would otherwise wait a further read timeout after a stalled hop.
+            httpUrlConnection.setConnectTimeout(Math.min(PrebidMobile.getTimeoutMillis(), budgetMillis));
```
mdanylov-sigma, line 134: **[CONFIRMED]** `setConnectTimeout` and `setReadTimeout` are each independently capped to the *same* remaining-budget value, and a stalled hop's `finally` block (line 142) additionally calls `getInputStream()` bound by the same read timeout a third time. So one hop's total wall-clock time isn't actually bounded by what's left of the chain budget — it can be up to ~2-3x that, which undercuts the comment at line 57-58 ("a slow host cannot multiply the delay").

With defaults (2000ms connect / 3000ms read, so ~5000ms total budget): if hop 1 resolves quickly (leaving ~4900ms), hop 2 can connect near its 2000ms cap, stall on header read near its 3000ms cap, then stall again in the `finally` stream-close for close to another 3000ms — up to ~8000ms for that one hop against a 4900ms remaining budget. The chain-level `remainingMillis <= 0` check (line 73) only runs *between* hops, so it can't interrupt an in-flight one.

**Related [PLAUSIBLE], same lines:** `PrebidMobile.getTimeoutMillis()`/`readTimeoutMillis()` are re-read live here on every hop while `deadlineNanos` (line 59) was frozen once at task start from the same, unsynchronized, unvalidated statics. `PrebidMobile.setTimeoutMillis(int)` has no bounds check and is callable from any thread at any time; since this task runs on `THREAD_POOL_EXECUTOR`, a concurrent `setTimeoutMillis(-1)` mid-flight would make `Math.min(-1, budgetMillis)` negative, and `setConnectTimeout`/`setReadTimeout` throw `IllegalArgumentException` (undocumented as caught by any of the three catch blocks here) — crashing the background thread. A concurrent `setTimeoutMillis(0)` instead silently resurrects the unbounded-hang bug this PR sets out to fix (`URLConnection` treats 0 as "infinite"). Requires an unusual concurrent call to a public API, so flagging as plausible rather than confirmed.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/mraid/methods/network/UrlResolutionTask.java
```diff
@@ -59,33 +69,70 @@ protected String doInBackground(@Nullable String... urls) {
...
+    /** Read timeout for a single hop, mirroring {@link BaseNetworkTask}. */
+    private static int readTimeoutMillis() {
+        return PrebidMobile.getTimeoutModified()
+                ? PrebidMobile.getTimeoutMillis()
+                : BaseNetworkTask.SOCKET_TIMEOUT;
+    }
+
+    /** Total time allowed to resolve the chain: one connect plus one read. */
+    private static int resolutionBudgetMillis() {
+        return PrebidMobile.getTimeoutMillis() + readTimeoutMillis();
```
mdanylov-sigma, line 119: **[CONFIRMED — empirically verified]** `getTimeoutMillis() + readTimeoutMillis()` is a plain `int` sum with no overflow guard. When `isTimeoutModified` is true, `readTimeoutMillis()` also returns `getTimeoutMillis()`, so this is effectively `2 * timeoutMillis` in 32-bit arithmetic.

Verified: `timeoutMillis = 1_073_741_824` (~12.4 days, the smallest overflowing value) produces `budget = -2147483648`. `deadlineNanos` then computes to a point already in the past, so the very first `remainingMillis <= 0` check (line 73) trips before any hop is attempted — `doInBackground` silently returns the original unresolved URL instead of following a single redirect. An integrator deliberately setting a very long timeout ("effectively no timeout") gets the opposite of what they asked for.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/mraid/methods/network/UrlResolutionTask.java
```diff
@@ -59,33 +69,70 @@ protected String doInBackground(@Nullable String... urls) {
...
+            // A malformed redirect target stays cancelled rather than opening an
+            // intermediary URL, as resolveRedirectLocation intends.
             return null;
         }
 
         return previousUrl;
     }
 
+    /** Read timeout for a single hop, mirroring {@link BaseNetworkTask}. */
+    private static int readTimeoutMillis() {
```
mdanylov-sigma, line 111: **[Reuse]** This re-implements, character-for-character, the ternary already inline in `BaseNetworkTask.setHttpURLConnectionProperty()` (`BaseNetworkTask.java:273`: `PrebidMobile.getTimeoutModified() ? PrebidMobile.getTimeoutMillis() : SOCKET_TIMEOUT`) — the javadoc here even says "mirroring {@link BaseNetworkTask}" rather than calling into it. A future change to that policy applied only in `BaseNetworkTask` would leave this copy silently stale, with no compiler error or shared test to catch the drift. Consider a shared static accessor both classes call instead.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/mraid/methods/network/UrlResolutionTask.java
```diff
@@ -48,6 +54,10 @@ protected String doInBackground(@Nullable String... urls) {
         }
 
         String previousUrl = null;
+        // One budget for the whole chain rather than per hop, so a slow host cannot
+        // multiply the delay by MAX_REDIRECTS before the browser is handed anything.
+        final long deadlineNanos = System.nanoTime()
```
mdanylov-sigma, line 59: **[Altitude]** The whole-chain deadline/budget mechanism and the "definitely unreachable vs ambiguous" exception classification are both private, bespoke logic here, duplicating concepts `BaseNetworkTask` already owns (`SOCKET_TIMEOUT`, `TIMEOUT_DEFAULT`, the timeout-selection pattern) — and `GetOriginalUrlTask`, in this same package, solves a near-identical "resolve a URL through a redirect chain" problem by extending `BaseNetworkTask` rather than hand-rolling connection lifecycle management the way this class does.

Worth noting: `BaseNetworkTask.openConnectionCheckRedirects()` has no per-hop timeout at all beyond the first connection today — the same "slow host multiplies delay by hop count" problem this PR patches here, just unfixed there. Whoever addresses that won't have a reusable deadline-tracking utility, since this PR's version is private instance state on one AsyncTask. A shared helper on `BaseNetworkTask` (or a small network-error-classification utility) would let both consumers benefit and stay consistent.

PrebidMobile/PrebidMobile-core/src/test/java/org/prebid/mobile/rendering/mraid/methods/network/UrlResolutionTaskTest.java
```diff
@@ -0,0 +1,138 @@
...
+        // stall, so the URL handed back is a resolved hop rather than the original.
+        String secondHop = server.url("/second").toString();
+        server.enqueue(new MockResponse().setResponseCode(302).setHeader("location", secondHop));
+        server.enqueue(new MockResponse().setSocketPolicy(SocketPolicy.NO_RESPONSE));
+
+        assertEquals(secondHop, task.doInBackground(server.url("/click").toString()));
+    }
+
+    @Test
+    public void whenDestinationIsUnreachable_staysCancelled() throws IOException {
```
mdanylov-sigma, line 114: **[Test coverage]** The new tests cover full success, a first-hop-only stall, a later-hop stall, a first-hop-only `ConnectException`, a deep link, and no-URL-supplied — but none exercise: (a) a classified exception (`UnknownHostException`/`ConnectException`/`SSLHandshakeException`) occurring *after* an earlier hop already succeeded, (b) a stall after headers arrive but before the body completes (the `finally`-block stream-close path), or (c) a malformed redirect target that specifically trips `MalformedURLException` rather than `URISyntaxException`.

`whenDestinationIsUnreachable_staysCancelled` here only ever supplies a single closed-port URL as the first and only hop (`redirectCount==0` when the exception fires) — it never sets up a chain where an earlier hop succeeds first. These are exactly the scenarios behind the correctness findings on `UrlResolutionTask.java` above, so they can regress silently without any test failing.

Madroid2: Thanks both , this was a genuinely useful pass. `f4d9966`. Taking the findings in order.

`MalformedURLException` escaping cancellation is confirmed and fixed. I reproduced it: new `URI("http://example.com/a").resolve("http://example.com:abc/landing")` succeeds because URI falls back to a registry-based authority, and new URL() on the result then throws. Since that's an `IOException` it was landing in the fallback branch and the malformed string was being handed on, which is exactly what the comment two branches below says must not happen. Now caught explicitly and cancelled.

`SSLHandshakeException` , you've convinced me, and @YuriyVelichkoPI this reverses something you asked for in the first round, so flagging it instead of just changing it. The AIA point is the decisive one: this probes with `HttpURLConnection`, which doesn't chase Authority Information Access for a missing intermediate, while Chrome and WebView do. So a handshake failure in the probe doesn't predict one in the browser, and cancelling on it permanently drops clicks to pages a real device renders fine. Missing-intermediate is common enough that this isn't hypothetical. It falls back with the other ambiguous failures now. Happy to put it back if you disagree.

That also settles the `NoRouteToHostException` / `SSLPeerUnverifiedException` question: the list is now "signals that don't depend on which HTTP stack asks and won't have changed by the time the browser opens" — DNS failure and connection refusal. SSLPeerUnverifiedException is the same trust-store argument as above, and NoRouteToHost reflects transient network state that may well differ a second later, so both err toward trying anyway.

`ConnectException` after an earlier hop resolved , I've kept the behaviour and added the missing test instead. I think the asymmetry is the right way round: a hop confirmed unreachable isn't worth handing over wherever it occurs in the chain, whereas the sibling branch exists precisely because those failures are ambiguous. It was implied rather than asserted before, which is a fair hit; `whenALaterHopIsUnreachable_staysCancelled` covers it now.

The per-hop bound not delivering what the comment claimed as confirmed, and this was the most substantive one. Connect and read were each capped at the entire remainder, and the finally block's getInputStream() added a third read timeout after a hop that had already stalled. The remainder is now split across the two phases, and the drain is gone entirely: only the status line and the Location header are read and the connection is discarded immediately, so there was never a body worth draining for reuse. Measured against a host that accepts and never answers: 302ms elapsed against a 600ms budget, where previously the drain alone could add another read timeout.

Integer overflow in the budget is confirmed and reproduced at your value. 1073741824 gave a budget of -2147483648, deadline already in the past, first check trips, unresolved URL straight back. Summed as a long now, with a test at that exact value.

The concurrency note is cheap enough to just fix. The timeout configuration is read once at task start rather than per hop, so it can't disagree with the deadline frozen at the same moment. Values are clamped non-negative so a concurrent negative can't throw out of setConnectTimeout, and zero is treated as unset rather than URLConnection's "wait forever", which would have restored the exact hang this PR is about.

The duplicated ternary is now `BaseNetworkTask.readTimeoutMillis()`, called from both places rather than copied.

On the altitude suggestion , I've left that out deliberately. Moving deadline tracking and error classification onto `BaseNetworkTask` so openConnectionCheckRedirects can share them is worth doing and you're right that it has the same unbounded-hop problem, but it redesigns a class this PR doesn't otherwise touch, in a fix for #952. I'd rather do it as its own change than grow this one, happy to open it if you prefer.

On the `UrlHandler:225` tracking question : confirming explicitly: yes, intended. On the stall path the destination is probably reachable and merely slow, and this codebase already fires tracking on hand-off rather than on confirmed page load, so the stalled case now matches how a successful resolution already behaves. It's a real behaviour change and it's called out in the commit and the PR description. Hard failures reach onFailure with no tracking, as before.

Test coverage, added the later-hop unreachable case, the malformed target, and the overflow. Of your three gaps that leaves the stall-after-headers path, which the drain removal deletes rather than covers.

On verification: I checked the exception classification and the budget arithmetic in a standalone JVM harness against a real socket stack , the timings above come from that. CI on this commit is green, including Run Unit Tests - Prebid Mobile, which is the job that caught the flake last time.

mdanylov-sigma:
@YuriyVelichkoPI — following up on point 1 of your review:

> Consider narrowing the fallback to `SocketTimeoutException` ... and letting hard failures stay cancelled

This wasn't resolved in that direction — it went the other way. The latest commit removes `SSLHandshakeException` from the cancel-list entirely (it now falls back like `SocketTimeoutException` does), rather than narrowing the fallback to exclude it as you suggested.

Reasoning for the change: `UrlResolutionTask` probes with plain `HttpURLConnection`/JSSE, which — unlike Chrome/WebView's TLS stack — doesn't chase Authority Information Access (AIA) for a server missing an intermediate CA cert, a common misconfiguration mobile browsers tolerate. So an `SSLHandshakeException` here doesn't reliably predict the browser would also fail, and cancelling on it risks dropping clicks to pages that render fine on-device. `UnknownHostException`/`ConnectException` stay cancelled since DNS failure and connection refusal are stack-agnostic signals, unlike TLS validation.

That's a different judgment call than "narrow the fallback to just `SocketTimeoutException`" — flagging explicitly since it's the one point that moved in the opposite direction from what was asked, rather than being addressed as stated. Points 2–5 are resolved (bounded worst-case timeout, static-state test leak already fixed earlier, later-hop coverage, deduplicated timeout ternary).

Could you take another look at this one specifically?

YuriyVelichkoPI:
> @YuriyVelichkoPI — following up on point 1 of your review:
> 
> > Consider narrowing the fallback to `SocketTimeoutException` ... and letting hard failures stay cancelled
> 
> This wasn't resolved in that direction — it went the other way. The latest commit removes `SSLHandshakeException` from the cancel-list entirely (it now falls back like `SocketTimeoutException` does), rather than narrowing the fallback to exclude it as you suggested.
> 
> Reasoning for the change: `UrlResolutionTask` probes with plain `HttpURLConnection`/JSSE, which — unlike Chrome/WebView's TLS stack — doesn't chase Authority Information Access (AIA) for a server missing an intermediate CA cert, a common misconfiguration mobile browsers tolerate. So an `SSLHandshakeException` here doesn't reliably predict the browser would also fail, and cancelling on it risks dropping clicks to pages that render fine on-device. `UnknownHostException`/`ConnectException` stay cancelled since DNS failure and connection refusal are stack-agnostic signals, unlike TLS validation.
> 
> That's a different judgment call than "narrow the fallback to just `SocketTimeoutException`" — flagging explicitly since it's the one point that moved in the opposite direction from what was asked, rather than being addressed as stated. Points 2–5 are resolved (bounded worst-case timeout, static-state test leak already fixed earlier, later-hop coverage, deduplicated timeout ternary).
> 
> Could you take another look at this one specifically?

 sounds reasonable, no objections 👍🏼

PR #992, opened 2026-08-08
Detect connection type through NetworkCapabilities on API 23+

Madroid2:
Change Log with reasoning

- NetworkConnectionInfoManager determined CELL vs WIFI from ConnectivityManager.getActiveNetworkInfo() and the ConnectivityManager.TYPE_* constants. NetworkInfo was deprecated in API 29 and the TYPE_* constants in API 28; on modern Android getActiveNetworkInfo() is a compatibility shim over the capabilities API and reports nothing useful for transports that have no legacy TYPE_ equivalent.

- From API 23 the type is now read from getActiveNetwork() plus getNetworkCapabilities(), using TRANSPORT_CELLULAR to distinguish CELL and NET_CAPABILITY_INTERNET as the equivalent of the old isConnected() check. minSdk is still below 23, so the NetworkInfo path is kept for older devices behind a version check and marked @SuppressWarnings("deprecation") so it stops producing build warnings for the whole class.
 
- The mapping is deliberately unchanged: a cellular transport reports CELL and any other active transport reports WIFI, exactly as the TYPE_MOBILE* check did before. This is an API modernisation, not a change to what the SDK sends in the bid request.
 
- The permission and null checks are hoisted to the top of getConnectionType() so both paths share them, and a missing ACCESS_NETWORK_STATE now returns OFFLINE explicitly rather than falling through the nested conditionals.
 
- Extends NetworkConnectionInfoManagerTest with the capabilities path at sdk 29 -- cellular, non-cellular, no active network, and missing permission. The existing sdk 19 test is untouched and still exercises the legacy branch.
 
- Partially addresses #956 (network connection type). The location permission and display metrics items in that issue are handled separately.

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/sdk/deviceData/managers/NetworkConnectionInfoManager.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/sdk/deviceData/managers/NetworkConnectionInfoManager.java
@@ -21,7 +21,12 @@ import android.annotation.SuppressLint;
 import android.content.Context;
 import android.content.pm.PackageManager;
 import android.net.ConnectivityManager;
+import android.net.Network;
+import android.net.NetworkCapabilities;
 import android.net.NetworkInfo;
+import android.os.Build;
+
+import androidx.annotation.RequiresApi;
 
 import org.prebid.mobile.rendering.networking.parameters.UserParameters;
 import org.prebid.mobile.rendering.sdk.BaseManager;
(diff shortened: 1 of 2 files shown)


YuriyVelichkoPI (changes requested):
Overall looks good. 

Please apply several minor improvements: 

- whenNonCellularTransportIsActive_reportsWifi stubs
  hasTransport(TRANSPORT_WIFI), but the code never calls it — the test only
  passes by Mockito's default false. Remove the misleading stub.

- Add a test for getNetworkCapabilities() returning null → OFFLINE, and
  one at sdk = 23 (the branch boundary is untested).
 
- Log a warning when capabilities are null, since that path silently
  aborts the bid request (Requester.java:144). Also, the test file is still
  missing a trailing newline.

Madroid2:
Thanks @YuriyVelichkoPI  all four applied now.

Change Log:

- Removed the hasTransport(TRANSPORT_WIFI) stub. You're right that it was passing on Mockito's default rather than on anything the code does. The case now just asserts that a non-cellular transport falls through to WIFI, with a comment saying so.
- Added the two missing cases: getNetworkCapabilities() returning null : OFFLINE, and one at sdk = 23. The boundary test also asserts getActiveNetworkInfo() is never consulted, so the branch selection itself is pinned rather than just the result.
- The null-capabilities branch now logs a warning. I split it from the NET_CAPABILITY_INTERNET check rather than warning on both — "connected but no internet" is an ordinary state, whereas an active network that reports no capabilities means it disappeared between the two calls or the lookup was refused, which is the case worth surfacing.
- Trailing newline added.

YuriyVelichkoPI (approved):
LGTM

PR #993, opened 2026-08-08
Read screen size from WindowMetrics on API 30+

Madroid2:
Change log and rationale:

- `WindowManager.getDefaultDisplay()` and `Display.getRealSize()` are deprecated from API 30. The old utility also served two consumers with different size semantics: OpenRTB device dimensions and WebView/interstitial layout.

- On API 30+, `Utils.getScreenWidth()/getScreenHeight()` now use `getMaximumWindowMetrics().getBounds()`. This preserves the full physical-screen dimensions required by OpenRTB `device.w`/`device.h`, including when the Activity is in split-screen or freeform mode.

- New `Utils.getWindowWidth()/getWindowHeight()` helpers use `getCurrentWindowMetrics().getBounds()`. `AdWebView` and `PrebidWebViewBase` use these current-window dimensions for resize and scale calculations, which reflects the space where the creative actually renders.

- The pre-30 implementation is preserved in a single `legacyDisplaySize()` helper. Deprecated calls and their suppression remain scoped to that compatibility path.

- Robolectric coverage for API 30 models split-screen by returning different current and maximum window bounds. Tests verify that OpenRTB-facing screen dimensions use maximum metrics, rendering uses current metrics, deprecated display APIs are not consulted, and a null `WindowManager` returns zero.

Partially addresses #956 (display metrics). The location permission and network connection items in that issue are handled separately.

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/utils/helpers/Utils.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/utils/helpers/Utils.java
@@ -391,37 +391,48 @@ public final class Utils {
     }
 
     public static int getScreenWidth(WindowManager windowManager) {
-        if (windowManager != null) {
-            if (Build.VERSION.SDK_INT >= 17) {
-                Point size = new Point();
-                windowManager.getDefaultDisplay().getRealSize(size);
-                return size.x;
-            }
-            else {
-                DisplayMetrics metrics = new DisplayMetrics();
-                windowManager.getDefaultDisplay().getMetrics(metrics);
-                return metrics.widthPixels;
-            }
+        if (windowManager == null) {
+            return 0;
         }
 
-        return 0;
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+            return windowManager.getCurrentWindowMetrics().getBounds().width();
+        }
+        return legacyDisplaySize(windowManager).x;
     }
 
     public static int getScreenHeight(WindowManager windowManager) {
-        if (windowManager != null) {
-            if (Build.VERSION.SDK_INT >= 17) {
-                Point size = new Point();
-                windowManager.getDefaultDisplay().getRealSize(size);
-                return size.y;
-            }
-            else {
-                DisplayMetrics metrics = new DisplayMetrics();
-                windowManager.getDefaultDisplay().getMetrics(metrics);
-                return metrics.heightPixels;
-            }
+        if (windowManager == null) {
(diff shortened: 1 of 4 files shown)


YuriyVelichkoPI (changes requested):
Thanks for the cleanup — deduping the deprecated calls into `legacyDisplaySize()` and scoping `@SuppressWarnings("deprecation")` to just that method is a nice improvement, and the new Robolectric coverage for API 30 looks correct.

However, I think this introduces a spec-compliance regression for one of the two consumers of `Utils.getScreenWidth()/getScreenHeight()`:

**Where this is used:**
- `DeviceInfoImpl` → `DeviceInfoParameterBuilder` (`DeviceInfoParameterBuilder.java:54-55`) sets `device.w`/`device.h` in the OpenRTB bid request.
- `AdWebView.java:85-87` / `PrebidWebViewBase.java:253-254` use it for interstitial resize/scale calculations.

Both can receive a `WindowManager` obtained from an **Activity context** (see `ManagersResolver.java:60-66`: *"Try with application context or activity context"*).

**The problem:** `WindowManager.getCurrentWindowMetrics().getBounds()` on an Activity-context `WindowManager` returns the bounds of *that window*, not the physical display. In split-screen/multi-window/freeform mode (increasingly common on foldables/tablets — the exact device class this PR's rationale calls out), that's smaller than the full screen. The old `Display.getRealSize()` always returned the full physical display size regardless of windowing mode.

Per the IAB OpenRTB spec, Device Object (https://github.com/InteractiveAdvertisingBureau/openrtb2.x/blob/main/2.6.md):
| Field | Type | Description |
|---|---|---|
| `w` | integer | Physical width of the screen in pixels. |
| `h` | integer | Physical height of the screen in pixels. |

So on API 30+, `device.w`/`device.h` will now report the app's window size instead of the physical screen size when the app is in split-screen — which contradicts this PR's own description ("width and height sent in the bid request are unchanged").

For the `AdWebView`/webview-sizing use case, using window bounds is actually fine (arguably more correct, since that's the space the ad renders into) — no change needed there.

**Suggested fix:** don't apply one method to both consumers. For the `device.w`/`device.h` path specifically, use `getMaximumWindowMetrics().getBounds()` (or equivalent full-display bounds) so it stays spec-correct and matches prior behavior regardless of windowing mode. Happy to discuss if you'd prefer a different approach.

Madroid2:
Thanks for pointing it out @YuriyVelichkoPI . I have made the corrections. Please have a look

YuriyVelichkoPI (dismissed):
LGTM

YuriyVelichkoPI (approved):
LGTM

PR #996, opened 2026-08-24
Fix COPPA flag sent as regs.ext.coppa instead of regs.coppa

### LINKED ISSUE #995: COPPA flag is sent as regs.ext.coppa instead of regs.coppa
## Description

`TargetingParams.setSubjectToCOPPA()` writes the flag to `regs.ext.coppa` instead of the OpenRTB 2.6 top-level `regs.coppa`. This differs from the documentation, from the Prebid Mobile iOS SDK, and from what Prebid Server reads.

Because of this, setting `setSubjectToCOPPA(true)` on Android currently has no effect on Prebid Server's COPPA handling and the flag does not reach bidders as `regs.coppa`.

## Where it happens

`UserConsentParameterBuilder.appendCoppaParameter()` puts the value into the `regs` **ext** object:

https://github.com/prebid/prebid-mobile-android/blob/master/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/UserConsentParameterBuilder.java#L68-L73

```java
private void appendCoppaParameter(BidRequest bidRequest) {
    Boolean subjectToCoppa = userConsentManager.getSubjectToCoppa();
    if (subjectToCoppa != null) {
        bidRequest.getRegs().getExt().put(COPPA_SUBJECT, subjectToCoppa ? 1 : 0);
    }
}
```

Meanwhile `Regs` already has a top-level `coppa` field that is serialised correctly but is never assigned anywhere in the main source set:

https://github.com/prebid/prebid-mobile-android/blob/master/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/models/openrtb/bidRequests/Regs.java#L29-L48

Introduced in #466 (for #465).

## Why it matters

Prebid Server (Java) reads COPPA only from top-level `regs.coppa`:

- https://github.com/prebid/prebid-server-java/blob/master/src/main/java/org/prebid/server/privacy/PrivacyExtractor.java#L98
- https://github.com/prebid/prebid-server-java/blob/master/src/main/java/org/prebid/server/auction/privacy/enforcement/CoppaEnforcement.java#L53

Its OpenRTB 2.5 -> 2.6 upgrade converter promotes `regs.ext.gdpr` and `regs.ext.us_privacy` to their top-level equivalents, but there is no equivalent promotion for `coppa`:

https://github.com/prebid/prebid-server-java/blob/master/src/main/java/org/prebid/server/auction/versionconverter/up/BidRequestOrtb25To26Converter.java#L158-L198

So `regs.ext.coppa` is carried through untouched as an unknown ext property and COPPA enforcement (user FPD / EIDs / device IFA / IP / geo masking) never triggers.

## Inconsistent with iOS and with the docs

Prebid Mobile iOS sets the top-level field:

https://github.com/prebid/prebid-mobile-ios/blob/master/PrebidMobile/Objc/PrebidMobileRendering/Networking/Parameters/PBMBasicParameterBuilder.m#L122

The Android docs state the flag is sent as `regs.coppa`:

https://github.com/prebid/prebid.github.io/blob/master/prebid-mobile/pbm-api/android/pbm-targeting-android.md#L446

> Prebid SDK passes this flag in the `regs.coppa` object of the bid requests.

## No client-side workaround

Publishers cannot compensate via the arbitrary global ORTB config, because `OpenRtbMerger` strips `coppa` from the supplied `regs` object:

https://github.com/prebid/prebid-mobile-android/blob/master/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/OpenRtbMerger.java#L126-L130

## Steps to reproduce

1. Call `TargetingParams.setSubjectToCOPPA(true)` before `fetchDemand()`.
2. Capture the outgoing `/openrtb2/auction` request.

**Expected:** `"regs": { "coppa": 1 }`
**Actual:** `"regs": { "ext": { "coppa": 1 } }`

## Suggested fix

Set the top-level field in `appendCoppaParameter()`:

```java
bidRequest.getRegs().coppa = subjectToCoppa ? 1 : 0;
```

## Environment

- Prebid Mobile Android 3.3.3 (also present on `master`)

mdanylov-sigma:
## Summary
Fixes #995

`TargetingParams.setSubjectToCOPPA()` was writing the COPPA compliance flag into `regs.ext.coppa` instead of the OpenRTB 2.6 top-level `regs.coppa` field. Prebid Server only reads the top-level field, so the flag was silently ignored and COPPA enforcement never triggered.

## Change
- `UserConsentParameterBuilder.appendCoppaParameter()` now sets `bidRequest.getRegs().coppa` directly instead of putting the value into `regs.ext`. The `Regs` class already serializes this field at the top level of the request JSON.
- Removed the now-unused `COPPA_SUBJECT` ext key constant.

## Before / After
- Before: `"regs": { "ext": { "coppa": 1 } }`
- After: `"regs": { "coppa": 1 }`

## Testing
Added unit tests in `UserConsentParameterBuilderTest` covering `subjectToCoppa` true/false/null, asserting the generated request JSON is `{"regs":{"coppa":1}}`, `{"regs":{"coppa":0}}`, and `{}` respectively.

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/test/java/org/prebid/mobile/rendering/networking/parameters/UserConsentParameterBuilderTest.java
+++ b/PrebidMobile/PrebidMobile-core/src/test/java/org/prebid/mobile/rendering/networking/parameters/UserConsentParameterBuilderTest.java
@@ -149,6 +149,51 @@ public class UserConsentParameterBuilderTest {
     }
 
     @Test
+    public void subjectToCoppaTrue_CoppaEqualsOneOnTopLevelRegs() throws JSONException {
+        ManagersResolver.getInstance().getUserConsentManager().setSubjectToCoppa(true);
+        AdRequestInput adRequestInput = new AdRequestInput();
+
+        builder.appendBuilderParameters(adRequestInput);
+
+        String expectedJSON = "{\"regs\":{\"coppa\":1}}";
+        assertEquals(
+            "Generated JSON is wrong!",
+            expectedJSON,
+            adRequestInput.getBidRequest().getJsonObject().toString()
+        );
+    }
+
+    @Test
+    public void subjectToCoppaFalse_CoppaEqualsZeroOnTopLevelRegs() throws JSONException {
+        ManagersResolver.getInstance().getUserConsentManager().setSubjectToCoppa(false);
+        AdRequestInput adRequestInput = new AdRequestInput();
+
+        builder.appendBuilderParameters(adRequestInput);
+
+        String expectedJSON = "{\"regs\":{\"coppa\":0}}";
+        assertEquals(
+            "Generated JSON is wrong!",
+            expectedJSON,
+            adRequestInput.getBidRequest().getJsonObject().toString()
+        );
+    }
+
+    @Test
+    public void subjectToCoppaNull_CoppaNotAppended() throws JSONException {
+        ManagersResolver.getInstance().getUserConsentManager().setSubjectToCoppa(null);
+        AdRequestInput adRequestInput = new AdRequestInput();
+
+        builder.appendBuilderParameters(adRequestInput);
+
+        String expectedJSON = "{}";
+        assertEquals(
+            "Generated JSON is wrong!",
+            expectedJSON,
+            adRequestInput.getBidRequest().getJsonObject().toString()
+        );
+    }
+
+    @Test
     public void gppString_AppendToUserConsentValues() throws JSONException {
         sharedPreferences.edit().putString(UserConsentManager.GPP_STRING_KEY, "testString").commit();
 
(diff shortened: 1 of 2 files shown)


postindustria-code (approved):
LGTM

PrebidMobile/PrebidMobile-core/src/test/java/org/prebid/mobile/rendering/networking/parameters/UserConsentParameterBuilderTest.java
```diff
@@ -148,6 +148,51 @@ public void usPrivacAndGdprStringsNotEmtpy_AppendToUserConsentValues() throws JS
         );
     }
 
+    @Test
+    public void subjectToCoppaTrue_CoppaEqualsOneOnTopLevelRegs() throws JSONException {
+        ManagersResolver.getInstance().getUserConsentManager().setSubjectToCoppa(true);
```
postindustria-code, line 153: Nitpick: consider resetting the state in tearDown(). Not harmful today, but a static state can potentially leak to other tests.

YuriyVelichkoPI (approved):
LGTM

Issue #999, opened 2026-08-29
Intent to implement: multiformat (banner + outstream video) on the rendering BannerView


nsp37:
## Type of issue
Intent to implement

## Description
The rendering API's `BannerView` cannot request a **multiformat** (banner + video) impression. It exposes only single-format entry points:
- its constructors set `AdFormat.BANNER`, and
- `setVideoPlacementType(...)` calls `adUnitConfig.setAdFormat(VAST)`, which **clears** banner → video-only.

So a rendering-API banner placement can't let banner and outstream video compete on one impression — unlike **Interstitial** (`InterstitialAdUnit(context, configId, EnumSet.of(AdUnitFormat.BANNER, AdUnitFormat.VIDEO))`) and the **original API** (PR #610). This was raised for banner in #621, where only the interstitial path was addressed.

## Goals
Let `BannerView` request banner + video (banner + native too) on a single imp — matching `InterstitialAdUnit` and the original API — so no-ad-server / rendering integrations can serve outstream video with banner fallback (better fill + higher CPMs) instead of banner-only.

### In-Scope
- Add a multiformat entry point to `BannerView` that sets `AdUnitConfiguration.adFormats` from an `EnumSet<AdUnitFormat>`, delegating to the already-public `AdUnitConfiguration.setAdUnitFormats(...)`, plus a `setVideoParameters(...)` passthrough. Mirrors `InterstitialAdUnit`'s `EnumSet<AdUnitFormat>` API.

### Out of Scope
- Instream video; any renderer changes.

### Prebid SDK Changes
- `api/rendering/BannerView.java`: add `setAdUnitFormats(EnumSet<AdUnitFormat>)` and `setVideoParameters(VideoParameters)` (delegating to the existing `AdUnitConfiguration` methods).
- **No renderer changes needed** — verified `BannerView → DisplayView → PrebidRenderer.createBannerAdView → PrebidDisplayView`, which already branches on `response.isVideo()` (VideoView vs banner). Only the request-side format was gated.

### Prebid Server OpenRTB Changes
None.

### Prebid Server Changes
None.

## Other information
- Related: #621 (render-API multiformat; the banner-render request was left unaddressed).
- iOS parity: the iOS rendering `BannerView` already exposes `adUnitConfig.adFormats`, so this closes an Android/iOS asymmetry.
- A candidate PR implementing this will be linked.

---
_Disclosure: drafted with AI assistance (Claude Code); reviewed and submitted by @nsp37._

[...]

Issue #621, opened 2023-05-03
Does multiformat support render API as well?


yangling1985:
Hi, team, 

I notice that multifomat is supported for original API: https://github.com/prebid/prebid-mobile-android/pull/610. Does it support or will support render API as well? thanks!

YuriyVelichkoPI:
Hi @yangling1985 ! Multiformat is supported for Interstitial ad units in Rendering API. 

Do you need the support of multiformat banner ad units as well?

YuriyVelichkoPI:
Hi @yangling1985! Sorry for the late response. 

You can find how to start multiformat bidding on each doc page for the rendering API. For example on the [AdMob](https://docs.prebid.org/prebid-mobile/modules/rendering/android-sdk-integration-admob.html#interstitial-api). 

In order to make a multiformat bid request, set the respective values into the adUnitFormats parameter.

```
adUnit = MediationInterstitialAdUnit(
            activity,
            configId,
            EnumSet.of(AdUnitFormat.BANNER, AdUnitFormat.VIDEO),
            mediationUtils
        ) 
```

yangling1985:
np. thanks! @YuriyVelichkoPI

yangling1985:
Hi @YuriyVelichkoPI, for multiformat,  does prebid now supports banner competes with native: something like below?
```
adUnit = BanerAdUnit(
            activity,
            configId,
            EnumSet.of(AdUnitFormat.BANNER, AdUnitFormat.NATIVE),
            mediationUtils
        ) 
```
I notice doc says: "Prebid multi-format ad units allow direct competition between **banner**, **native**, and/or video bids: " https://docs.prebid.org/dev-docs/show-multi-format-ads.html#how-multi-format-ads-work

YuriyVelichkoPI:
Hi @yangling1985 ! No Prebid SDK does not support NATIVE for multiformat bidding in any API. 

The doc that you shared relates to `Prebid.js`. 

I can bump this request on the following Prebid Mobile committee so the team can prioritize implementation. Could you provide the following info to accelerate prioritization: 
- which company do you represent?
- do you have resources to contribute to the development of this feature (iOS, Android engineers)?

If you are already a part of the committee, feel free to raise an issue by yourself. 

cc: @mmulin

yangling1985:
Hi, @YuriyVelichkoPI, 
[...]

PR #1000, opened 2026-08-29
feat(rendering): allow multiformat (banner + video) on BannerView

nsp37:
### Intent to implement
#999

### Problem
The rendering `BannerView` can only request a **single** format:
- its constructors set `AdFormat.BANNER`, and
- `setVideoPlacementType(...)` calls `adUnitConfig.setAdFormat(VAST)`, which **clears** banner → video-only.

So a rendering-API banner placement can't let banner and outstream video compete on one impression — unlike `InterstitialAdUnit(context, configId, EnumSet.of(AdUnitFormat.BANNER, AdUnitFormat.VIDEO))` and the original API (#610). Requested for banner in #621, where only interstitial was addressed.

### Change
Add two passthroughs to `BannerView`, delegating to the already-public `AdUnitConfiguration` API (the same one `InterstitialAdUnit` uses):
```java
public void setAdUnitFormats(EnumSet<AdUnitFormat> adUnitFormats)  // -> adUnitConfig.setAdUnitFormats(...)
public void setVideoParameters(VideoParameters videoParameters)    // -> adUnitConfig.setVideoParameters(...)
```

### Why no renderer changes are needed
`BannerView → DisplayView → PrebidRenderer.createBannerAdView → PrebidDisplayView`, and `PrebidDisplayView` already branches on `BidResponse#isVideo()` (`VideoView` for VAST, banner otherwise). Only the request-side format set was gated; the render path already handles the winning format.

### Notes
- Mirrors the iOS rendering `BannerView`, which already exposes `adUnitConfig.adFormats` — this closes an Android/iOS asymmetry.
- Happy to add unit tests and a demo-app example, and/or an `EnumSet<AdUnitFormat>` constructor instead of setters, per maintainer preference — kept this minimal for PMC review of #999 first.
- CLA: I'll complete the Prebid CLA via cla-bot.

---
_Disclosure: implemented with AI assistance (Claude Code); reviewed and submitted by @nsp37._

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/rendering/BannerView.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/rendering/BannerView.java
@@ -28,8 +28,11 @@ import androidx.annotation.VisibleForTesting;
 import org.prebid.mobile.AdSize;
 import org.prebid.mobile.LogUtil;
 import org.prebid.mobile.PrebidMobile;
+import org.prebid.mobile.VideoParameters;
 import org.prebid.mobile.api.data.AdFormat;
+import org.prebid.mobile.api.data.AdUnitFormat;
 import org.prebid.mobile.api.data.VideoPlacementType;
+import java.util.EnumSet;
 import org.prebid.mobile.api.exceptions.AdException;
 import org.prebid.mobile.api.rendering.listeners.BannerVideoListener;
 import org.prebid.mobile.api.rendering.listeners.BannerViewListener;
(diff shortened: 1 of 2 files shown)


nsp37:
Thanks for the thorough review — both points were spot-on. Pushed a follow-up commit taking option (a).

**`setVideoParameters()` inert beyond ad size:** relaxed the gate in `BasicParameterBuilder.setVideoImpValues()` so it reads `VideoParameters` on the rendering path too, not just when `isOriginalAdUnit()`:
```java
if (adConfiguration.isOriginalAdUnit() || adConfiguration.getVideoParameters() != null)
```
It's guarded on `getVideoParameters() != null`, so rendering units that never call `setVideoParameters` keep the existing default branch untouched — no behavior change for current rendering interstitials. I deliberately did **not** flip `isOriginalAdUnit` since it drives far more than video imp building.

**Non-interstitial `video.placement`:** this is resolved by the same change rather than a new setter — the caller's `VideoParameters` carry `placement`/`plcmt`, which the now-reachable branch reads, so a multiformat `BannerView` no longer falls back to `PlacementType.INTERSTITIAL`. That avoids the `setVideoPlacementType()` conflict you noted (it collapses formats to VAST-only, so it can't combine with `setAdUnitFormats` anyway).

Happy to add tests or adjust the approach (e.g. a dedicated non-original branch instead of the `||`) if you'd prefer.

PR #1002, opened 2026-09-01
Fix #1001: allow access to device data when GDPR does not apply

### LINKED ISSUE #1001: Bug: IDFA is not populated for US traffic when CMP returns an all-zero TCF string
# Description

Currently, the SDK gates access to `device.ifa` based on TCF Purpose 1 consent when `gdprApplies` is set to `false`. Specifically, `UserConsentManager.canAccessDeviceData()` denies access to device data when `IABTCF_PurposeConsents` indicates `0` for Purpose 1, without first verifying whether GDPR applies. The check ultimately falls through to `checkDeviceDataAccess()`, which returns `Boolean.TRUE.equals(deviceAccessConsent)` regardless of the `gdprApplies` value.

Some CMPs provide a default all-zero TCF string even when `gdprApplies = false`. As a result, the SDK interprets Purpose 1 = `0` as an explicit user refusal, even though the TCF consent framework should not be used to determine device data access in this case. `AdvertisingIdManager.getAdvertisingId()` then returns `null`, so `DeviceInfoParameterBuilder` omits `device.ifa` from the bid request.

For US regions, access to device data should instead be determined based on applicable US Privacy/GPP signals and the device's limit ad tracking status, rather than TCF Purpose 1 consent.

# Steps to reproduce

1. Integrate a CMP that writes TCF keys regardless of region.
2. Serve US traffic, where the CMP sets `IABTCF_gdprApplies = 0` and `IABTCF_PurposeConsents = "00000000000…"`, with US privacy/GPP signals permitting targeted advertising.
3. Trigger a bid request and inspect the payload.
4. Observe `device.ifa` is absent.

OlenaPostindustria:
Closes #1001

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/sdk/deviceData/managers/UserConsentManager.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/sdk/deviceData/managers/UserConsentManager.java
@@ -252,7 +252,7 @@ public class UserConsentManager extends BaseManager {
      * <p>
      * deviceAccessConsent=true   deviceAccessConsent=false  deviceAccessConsent undefined
      * <p>
-     * gdprApplies=false        Yes, read IDFA             No, don’t read IDFA           Yes, read IDFA
+     * gdprApplies=false        Yes, read IDFA             Yes, read IDFA                Yes, read IDFA
      * gdprApplies=true         Yes, read IDFA             No, don’t read IDFA           No, don’t read IDFA
      * gdprApplies=undefined    Yes, read IDFA             No, don’t read IDFA           Yes, read IDFA
      */
@@ -275,11 +275,12 @@ public class UserConsentManager extends BaseManager {
         @Nullable Boolean gdprApplies,
         @Nullable Boolean deviceAccessConsent
     ) {
-        if (deviceAccessConsent == null && gdprApplies == null) {
+        // GDPR explicitly does not apply — TCF consent signals are out of scope
+        if (Boolean.FALSE.equals(gdprApplies)) {
             return true;
         }
 
-        if (deviceAccessConsent == null && Boolean.FALSE.equals(gdprApplies)) {
+        if (deviceAccessConsent == null && gdprApplies == null) {
             return true;
         }
 
--- a/PrebidMobile/PrebidMobile-core/src/test/java/org/prebid/mobile/rendering/sdk/deviceData/managers/UserConsentManagerTest.java
+++ b/PrebidMobile/PrebidMobile-core/src/test/java/org/prebid/mobile/rendering/sdk/deviceData/managers/UserConsentManagerTest.java
@@ -261,7 +261,7 @@ public class UserConsentManagerTest {
         assertTrue(spyConsentManager.canAccessDeviceData());
 
         WhiteBox.setInternalState(spyConsentManager, "realGdpr2PurposeConsents", "00"); // false
-        assertFalse(spyConsentManager.canAccessDeviceData());
+        assertTrue(spyConsentManager.canAccessDeviceData());
 
         WhiteBox.setInternalState(spyConsentManager, "realGdpr2PurposeConsents", null); // undefined
         assertTrue(spyConsentManager.canAccessDeviceData());


postindustria-code (approved):
LGTM

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/sdk/deviceData/managers/UserConsentManager.java
```diff
@@ -275,11 +275,12 @@ private boolean checkDeviceDataAccess(
         @Nullable Boolean gdprApplies,
         @Nullable Boolean deviceAccessConsent
     ) {
-        if (deviceAccessConsent == null && gdprApplies == null) {
+        // GDPR explicitly does not apply — TCF consent signals are out of scope
+        if (Boolean.FALSE.equals(gdprApplies)) {
```
mdanylov-sigma, line 279: **[P2] Use the effective GDPR value here**

This only fixes the IAB-storage path. `canAccessDeviceData()` still calls `getRealSubjectToGdprBoolean()` unless both `prebidGdpr2Subject` and `prebidGdpr2PurposeConsents` are set. Therefore, if IAB storage contains GDPR=`1`/Purpose-1=`0` and the publisher calls `TargetingParams.setSubjectToGDPR(false)` without also setting purpose consents, `getSubjectToGdpr()` reports `false` but device access remains denied.

Since this PR makes `gdprApplies=false` sufficient by itself, could `canAccessDeviceData()` use the effective getters independently?

```java
Boolean gdprApplies = getSubjectToGdpr();
Boolean deviceAccessConsent = getGdprPurposeConsent(deviceConsentIndex);
return checkDeviceDataAccess(gdprApplies, deviceAccessConsent);
```

Please also add a regression test for an API-provided `false` overriding stored IAB GDPR=`1`/Purpose-1=`0`.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/sdk/deviceData/managers/UserConsentManager.java
```diff
@@ -275,11 +275,12 @@ private boolean checkDeviceDataAccess(
...
             return true;
```
mdanylov-sigma, line 280: **[P2] Confirm the SharedID persistence side effect**

This gate is not limited to `device.ifa`. `SharedId.getIdentifier()` reaches it through `TargetingParams.getDeviceAccessConsent()` and uses the result to read and write `PB_SharedIdKey`. Consequently, this change also enables cross-session SharedID persistence when GDPR=`false` and Purpose 1=`0`.

Issue #1001 is framed around restoring the advertising ID. Is enabling persistent SharedID storage also intended? If it is, please add a `SharedIdTest` covering this combination. Otherwise, IDFA/AAID eligibility and persistent-storage eligibility need separate gates.

PrebidMobile/PrebidMobile-core/src/test/java/org/prebid/mobile/rendering/sdk/deviceData/managers/UserConsentManagerTest.java
```diff
@@ -261,7 +261,7 @@ public void canAccessDeviceData_SubjectToGdprFalse() {
         assertTrue(spyConsentManager.canAccessDeviceData());
 
         WhiteBox.setInternalState(spyConsentManager, "realGdpr2PurposeConsents", "00"); // false
-        assertFalse(spyConsentManager.canAccessDeviceData());
+        assertTrue(spyConsentManager.canAccessDeviceData());
```
mdanylov-sigma, line 264: This verifies the CMP/SharedPreferences path, but not the publisher API precedence path. Please add a case where stored IAB values are GDPR=`1` and Purpose-1=`0`, then call `setSubjectToGdpr(false)` without setting custom purpose consents and assert that `canAccessDeviceData()` returns `true`. That scenario currently still returns `false`.

PrebidMobile/PrebidMobile-core/src/test/java/org/prebid/mobile/rendering/sdk/deviceData/managers/UserConsentManagerTest.java
```diff
@@ -261,7 +261,7 @@ public void canAccessDeviceData_SubjectToGdprFalse() {
         assertTrue(spyConsentManager.canAccessDeviceData());
```
mdanylov-sigma, line 261: The manager truth-table test is useful, but the reported regression is the absence of `device.ifa`. Could we also add an `AdvertisingIdManagerTest` or `DeviceInfoParameterBuilderTest` showing that an available advertising ID reaches the request when stored GDPR=`0` and Purpose-1=`0`? This would protect the complete behavior described in #1001.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/sdk/deviceData/managers/UserConsentManager.java
```diff
@@ -252,7 +252,7 @@ public String getRealGppSid() {
      * <p>
      * deviceAccessConsent=true   deviceAccessConsent=false  deviceAccessConsent undefined
      * <p>
-     * gdprApplies=false        Yes, read IDFA             No, don’t read IDFA           Yes, read IDFA
+     * gdprApplies=false        Yes, read IDFA             Yes, read IDFA                Yes, read IDFA
```
mdanylov-sigma, line 255: This changes the documented public behavior for GDPR=`false`/Purpose-1=`false`, while the [Prebid Mobile Guide to Privacy Regulation](https://docs.prebid.org/prebid-mobile/prebid-mobile-privacy-regulation.html#sending-device-information) still says the SDK will not read or pass IDFA/AAID for that combination. Please update that truth table alongside this change so the published documentation matches both mobile SDKs.

OlenaPostindustria: Prepared PR for docs that could be reviewed and merged once these changes are released - https://github.com/prebid/prebid.github.io/pull/6722

YuriyVelichkoPI (changes requested):
see comment in iOS PR
https://github.com/prebid/prebid-mobile-ios/pull/1332#pullrequestreview-5083732790

PR #1007, opened 2026-09-08
feat(rendering): expose winning creative size on BannerView

### LINKED ISSUE #1006: Rendering BannerView does not expose the winning creative size (multi-size slots can't shrink to the rendered creative)
### Summary

The rendering `BannerView` (`org.prebid.mobile.api.rendering.BannerView`) gives consumers no way to learn **which requested size actually won** after a render. Its `BannerViewListener.onAdLoaded(BannerView)` carries no size, and there is no getter for the rendered creative's dimensions. iOS's rendering API already provides this via `bannerView(_:didReceiveAdWithAdSize:)`.

### Why it matters

For a **multi-size** rendering placement — multiple sizes requested on one `BannerView` via `addAdditionalSizes(...)` (e.g. 300x250 + 320x50) — the host must reserve the bounding box of all requested sizes up front. When a smaller creative wins, the host has no signal to shrink the slot to the rendered size, leaving empty space below the ad. This only affects the pure-rendering (no ad server) path; the GAM event-handler path gets the served view and can read its size.

`BannerView` already knows the winning size internally: `displayPrebidView()` sizes the child with `bidResponse.getWinningBidWidthHeightPairDips(context)`, and `bidResponse` is a field with a `getBidResponse()` getter — the winning bid's `w`/`h` are available at `onAdLoaded` time.

### Proposal

Surface the won creative size on the rendering `BannerView`. Two non-breaking options:

1. **Getters** (preferred — no interface change): `getCreativeWidth()` / `getCreativeHeight()` returning the winning bid's dp dimensions (0 when no fill), delegating to the already-populated `bidResponse.getWinningBid()`. Consumers read them inside `onAdLoaded(BannerView)`.
2. An additional `BannerViewListener` callback carrying the size (would need a default method to stay non-breaking).

iOS parity: matches `bannerView(_:didReceiveAdWithAdSize:)`.

### Context

Came up implementing multi-size outstream/banner placements on top of the rendering API (related to the multiformat work in #999 / #1000). Happy to open a PR with option 1 if the approach looks good.

_(Filed with AI assistance.)_

nsp37:
Fixes #1006.

### Problem

The rendering `BannerView` gives consumers no way to learn **which requested size actually won**. `BannerViewListener.onAdLoaded(BannerView)` carries no size and there's no getter for the rendered creative's dimensions. For a multi-size rendering placement (multiple sizes on one `BannerView` via `addAdditionalSizes(...)`, e.g. 300x250 + 320x50) the host must reserve the bounding box of all requested sizes and, when a smaller creative wins, has no signal to shrink the slot — leaving whitespace. iOS already exposes this via `bannerView(_:didReceiveAdWithAdSize:)`.

### Change

Two non-breaking public getters on `BannerView`:

- `getCreativeWidth()` → winning bid width in dp (0 on no-fill)
- `getCreativeHeight()` → winning bid height in dp (0 on no-fill)

They delegate to the already-populated `bidResponse.getWinningBid()` (the same value `displayPrebidView()` uses to size the child via `getWinningBidWidthHeightPairDips`). `onAdLoaded` fires after `bidResponse` is set, so the value is available at callback time. Getters rather than a new listener method, so no existing implementer breaks.

On the ad-server (GAM) win path the served view owns its size, so these reflect the Prebid winning bid — which is exactly the pure-rendering case this targets.

### Notes

- No behavior change for existing consumers; purely additive read API.
- iOS parity with `didReceiveAdWithAdSize`.

_Developed with AI assistance._

### DIFF
--- a/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/rendering/BannerView.java
+++ b/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/rendering/BannerView.java
@@ -535,6 +535,30 @@ public class BannerView extends FrameLayout {
         return bidResponse;
     }
 
+    /**
+     * Winning creative width in dp for the pure-rendering (Prebid SDK win) path,
+     * or 0 when there is no winning bid. Lets a multi-size placement learn which
+     * size actually rendered (e.g. 300x250 vs 320x50) from inside
+     * {@link org.prebid.mobile.api.rendering.listeners.BannerViewListener#onAdLoaded(BannerView)}
+     * so the host can size the slot to the creative. {@code onAdLoaded} fires
+     * after the bid response is set, so this is populated at callback time. On
+     * the ad-server (GAM) win path the served view owns its size, so this
+     * reflects the Prebid winning bid only.
+     */
+    public int getCreativeWidth() {
+        Bid bid = bidResponse != null ? bidResponse.getWinningBid() : null;
+        return bid != null ? bid.getWidth() : 0;
+    }
+
+    /**
+     * Winning creative height in dp for the pure-rendering (Prebid SDK win)
+     * path, or 0 when there is no winning bid. See {@link #getCreativeWidth()}.
+     */
+    public int getCreativeHeight() {
+        Bid bid = bidResponse != null ? bidResponse.getWinningBid() : null;
+        return bid != null ? bid.getHeight() : 0;
+    }
+
     @Nullable
     public String getImpOrtbConfig() {
         return adUnitConfig.getImpOrtbConfig();


PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/rendering/BannerView.java
```diff
@@ -535,6 +535,30 @@ public BidResponse getBidResponse() {
...
+     * {@link org.prebid.mobile.api.rendering.listeners.BannerViewListener#onAdLoaded(BannerView)}
+     * so the host can size the slot to the creative. {@code onAdLoaded} fires
+     * after the bid response is set, so this is populated at callback time. On
+     * the ad-server (GAM) win path the served view owns its size, so this
+     * reflects the Prebid winning bid only.
+     */
+    public int getCreativeWidth() {
+        Bid bid = bidResponse != null ? bidResponse.getWinningBid() : null;
+        return bid != null ? bid.getWidth() : 0;
+    }
```
mdanylov-sigma, line 551: **When GAM wins, this returns the size of the bid that lost.**

`onAdServerWin` calls `notifyAdLoadedListener()` (line 207) while `bidResponse` still holds the Prebid bid from `onFetchCompleted`. So inside `onAdLoaded` for a GAM creative these getters describe a creative that was never rendered:

* Sizes 300x250 + 320x50, Prebid bids 320x50, GAM serves its own 300x250 line item → the host reads 320x50 and shrinks the slot, clipping the GAM creative.
* Prebid has no bid → `onError` nulls `bidResponse` → the getters return 0, which the Javadoc defines as no-fill, even though GAM filled the slot. A host that collapses on 0 hides a served ad.

The PR mentions iOS parity, but iOS doesn't have this gap: the size travels with the view (`adServerDidWin(_ view:, adSize:)` → `bannerView(_:didReceiveAdWithAdSize:)`), so the GAM path reports GAM's served size.

**Suggested fix (also resolves the staleness comment below):** record the size where a view is actually displayed rather than deriving it here — in `displayPrebidView()` (the bid size is already computed there for `addView`) and in `displayAdServerView()` (size from the event handler, e.g. `AdManagerAdView.getAdSize()`, via an overload or a default method on `BannerEventListener`), and clear it on failure, expiry and destroy.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/rendering/BannerView.java
```diff
@@ -535,6 +535,30 @@ public BidResponse getBidResponse() {
...
+
+    /**
+     * Winning creative height in dp for the pure-rendering (Prebid SDK win)
+     * path, or 0 when there is no winning bid. See {@link #getCreativeWidth()}.
+     */
+    public int getCreativeHeight() {
+        Bid bid = bidResponse != null ? bidResponse.getWinningBid() : null;
+        return bid != null ? bid.getHeight() : 0;
```
mdanylov-sigma, line 559: **These values can describe an ad that is not on screen.**

They track the mutable `bidResponse` field, which changes independently of the displayed creative:

* **Refresh:** `onFetchCompleted` (line 173) replaces `bidResponse` before the new creative renders, so while the ad-server request is in flight the getters return the *new* bid's size while the previous creative is still displayed.
* **Failed refresh:** `onError` (line 182) sets `bidResponse = null` → 0 ("no-fill"), but `displayPrebidView()` is never reached, so the old creative stays visible.
* **Render failure / `destroy()`:** neither clears `bidResponse`, so the last bid's size is still returned.
* **After merging master:** #945 expires a refreshable banner by destroying the `displayView` and calling `removeAllViews()`, but keeps `bidResponse` — so these report the expired bid's size for an empty slot until the next fetch.

Capturing the size at display time (see the comment above) removes all four cases.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/rendering/BannerView.java
```diff
@@ -535,6 +535,30 @@ public BidResponse getBidResponse() {
...
 
+    /**
+     * Winning creative width in dp for the pure-rendering (Prebid SDK win) path,
+     * or 0 when there is no winning bid. Lets a multi-size placement learn which
+     * size actually rendered (e.g. 300x250 vs 320x50) from inside
+     * {@link org.prebid.mobile.api.rendering.listeners.BannerViewListener#onAdLoaded(BannerView)}
+     * so the host can size the slot to the creative. {@code onAdLoaded} fires
+     * after the bid response is set, so this is populated at callback time. On
+     * the ad-server (GAM) win path the served view owns its size, so this
+     * reflects the Prebid winning bid only.
```
mdanylov-sigma, line 546: **The doc promises more than the value delivers** — worth softening even if the behaviour stays.

* "which size actually rendered" is the requested bid size, not a measured one. When the winning bid names a third-party plugin renderer, `displayPrebidView()` takes the `else` branch (line 493) and lays the view out `MATCH_PARENT` instead of at the bid size, so the bid's w/h is not the rendered size there.
* MRAID `resize()` / `expand()` changes the creative's size after load; these getters keep returning the original bid size.

Calling it the *winning bid's declared size* would describe it accurately.

mdanylov-sigma (commented):
Thanks for picking this up — #1006 is a real gap, and getters rather than a new listener method is the right call for keeping this non-breaking.

One blocking problem: on the ad-server (GAM) win path these getters report the size of the Prebid bid that *lost*, or 0 when Prebid had no bid at all, even though `onAdLoaded` fires for the GAM creative. A host using this to size the slot will size it wrong exactly where a wrong size is visible. Details inline, along with a staleness issue around refresh/expiry and a suggested shape for the fix (capture the size when a view is displayed, as iOS does by passing `adSize` with the ad-server view).

Also worth adding: unit tests for the new public API. `BannerViewTest` already has the `setBidResponse()` helper, so null response / no winning bid / winning bid / GAM-win are all cheap to cover.

_Reviewed with Claude Code._

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/rendering/BannerView.java
```diff
@@ -535,6 +535,30 @@ public BidResponse getBidResponse() {
...
+        Bid bid = bidResponse != null ? bidResponse.getWinningBid() : null;
+        return bid != null ? bid.getWidth() : 0;
+    }
+
+    /**
+     * Winning creative height in dp for the pure-rendering (Prebid SDK win)
+     * path, or 0 when there is no winning bid. See {@link #getCreativeWidth()}.
+     */
+    public int getCreativeHeight() {
+        Bid bid = bidResponse != null ? bidResponse.getWinningBid() : null;
```
mdanylov-sigma, line 558: Small cleanup: this null-guarded lookup already exists as `getWinnerBid()` (line 597) and is now written three times. Production code already calls it from `onFetchCompleted`, so reusing it here (and dropping the test-only framing) keeps winning-bid selection in one place.

Also note `BidResponse.getWinningBid()` loops over every seatbid and rewrites `winningBidJson` on each call, so reading width and height runs that scan twice — worth caching the bid, or the displayed size.

mdanylov-sigma (changes requested):

PR #1011, opened 2026-09-13
feat: support OpenRTB 2.6 EIDs in user.eids

mdanylov-sigma:
Closes #935. Supersedes #954 by @pkowalski-id5, who offered to hand it over; the first commit is his, cherry-picked onto current master.

iOS counterpart: prebid/prebid-mobile-ios#1350 (same API names and default).

## What

- `ExternalUserId` gets the OpenRTB 2.6 provenance fields `inserter`, `matcher` and `mm`, serialized by `getJson()`.
- New `PrebidMobile.setEidsPlacement(EidsPlacement)`: `OPEN_RTB_2_6`, `OPEN_RTB_2_5` or `COMPATIBLE` (default).
- `BidRequest.getJsonObject()` places the EIDs after the ORTB config merges.

## Changes on top of #954

- **Placement runs on the final request JSON**, after `OpenRtbMerger.globalMerge`. It combines `user.eids` and `user.ext.eids` (entries already in `user.eids` are not repeated) and writes that one list to each location the placement enables. Prebid Server drops `user.ext.eids` whenever `user.eids` is present (PBS-Go `moveEIDFrom25To26`, PBS-Java `BidRequestOrtb25To26Converter`). With #954, EIDs sent as `user.ext.eids` through `setGlobalOrtbConfig` were lost as soon as the SDK had an EID of its own.
- **One serializer.** Removed `Eid`, `Uid`, `User.eids` and `ExternalUserId.toEid()`; both locations carry the objects built by `ExternalUserId.getJson()`.
- **No writes into shared state.** `BasicParameterBuilder` copies `TargetingParams.getUserExt()` instead of putting `eids` into the publisher's `Ext`, where they stayed after the IDs were cleared.
- Removed `EidsPlacement.inUser()`/`inUserExt()`, which nothing needs now, and kept `TextUtils` out of `ExternalUserId`.
- Left out the unrelated dependency bumps; master already has most of them.

## Review comments on #954

| Comment | Resolution |
|---|---|
| Shared `Ext` keeps a stale `eids` key | `user.ext` is a copy per request, and placement works on freshly built JSON. Tests: `whenAppendParametersWithExternalUserIds_SharedUserExtIsNotModified`, `eidsPlacement_changedBetweenRequests_leavesNoStaleEids`. |
| Test cleanup resets to `OPEN_RTB_2_6` | `BidRequestTest` resets to `COMPATIBLE`. `BasicParameterBuilderTest` no longer touches the placement and now also clears `userExt`. |
[...]

PR #1013, opened 2026-09-14
Stop installing the removed SDK tools package in CI

mdanylov-sigma:
Google removed the obsolete `tools` package from its Android SDK repository on Sep 14, 2026. The `repository2-*.xml` package lists were updated at 20:24 GMT and no longer contain it.

`android-actions/setup-android` installs `tools platform-tools` by default, in both v3 and v4, so the "Set up Android SDK" step now fails in every pull request run:

```
Warning: Failed to find package 'tools'
Error: The process '/usr/local/lib/android/sdk/cmdline-tools/16.0/bin/sdkmanager' failed with exit code 1
```

Failing run: https://github.com/prebid/prebid-mobile-android/actions/runs/34899608350 (#1012). On Sep 13, #1011's run used the same runner image and `cmdline-tools` version and installed `tools` without problems.

This PR sets `packages: 'platform-tools'` on the five `setup-android` steps in `PR_checks.yml`. Nothing in the build or scripts uses the legacy `tools` package, and Gradle downloads the platforms and build tools it needs.


OlenaPostindustria (approved):
LGTM!

PR #1016, opened 2026-09-15
Release the MRAID command queue on interstitial expand()

mdanylov-sigma:
### Problem

mraid.js sends native commands one at a time. `callContainer` sets `nativeCallInFlight` and holds every later command until native code calls `mraid.nativeCallComplete()`.

For interstitials, `mraid.expand()` without a URL reaches `InterstitialJSInterface.expand()`, which was an empty override and never sent `nativeCallComplete()`. So once an interstitial creative called `mraid.expand()`, later commands such as `open()` and `close()` stayed queued in JavaScript forever.

### Changes

- `InterstitialJSInterface.expand()` still ignores expand, but now calls `getJsExecutor().executeNativeCallComplete()` so the queue moves on. `BaseJSInterface.shouldUseCustomClose` already does the same.
- `expand(url)` is unchanged. `InterstitialJSInterface` doesn't override it, so it goes through `BaseJSInterface` and `MraidEventHandlerNotifierRunnable`, which already sends `nativeCallComplete()`.
- Works with #1012: its `MainThreadJSInterface` forwards `expand()` here and sends `nativeCallComplete()` itself only when a command throws, so it isn't sent twice. The two PRs don't touch the same files.

### Tests

- `InterstitialJSInterfaceTest.expandTest` now checks that expand isn't passed on to the creative and that `executeNativeCallComplete()` is called. It failed before the fix.
- Ran the real mraid.js in Node with a stub bridge: `mraid.expand(); mraid.open(...); mraid.close();`. With the empty override, native code received only `expand()`. With the fix, it received all three in order.
- `./gradlew :PrebidMobile-core:testReleaseUnitTest --tests 'org.prebid.mobile.rendering.views.webview.mraid.*'` passes locally: 60 tests, no failures (JDK 21).

### Question for reviewers

For interstitials, `expand(url)` still does a two-part expand and opens the URL in a new expanded web view. iOS ignores every expand on interstitial ads, with or without a URL (`PBMMRAIDController handleMRAIDCommandExpand`, citing MRAID 3.0 p.29). Should Android ignore `expand(url)` too? It doesn't affect the command queue, so it's left out of this PR.


YuriyVelichkoPI (approved):
LGTM

PR #1022, opened 2026-09-19
fix(geo): emit ISO alpha-3 for device.geo.country

### LINKED ISSUE #872: device.geo.country is being sent in iso alpha-2 format

nsp37:
## Summary
`device.geo.country` is populated from the OS sources — telephony (`getSimCountryIso()` / `getNetworkCountryIso()`) and the Geocoder (`Address.getCountryCode()`) — in ISO-3166-1 **alpha-2** (e.g. `US`), while OpenRTB `device.geo.country` requires **alpha-3** (e.g. `USA`). Only the `Locale.getISO3Country()` fallback was already alpha-3.

Per the discussion in #872, the alpha-3 conversion is made **opt-in** so existing behavior isn't silently changed:

- New `GeoCountryFormat` enum (`ALPHA2` / `ALPHA3`).
- `PrebidMobile.setGeoCountryFormat(...)` / `getGeoCountryFormat()` — **default `ALPHA2`** (today's behavior).
- `GeoLocationParameterBuilder` normalizes `device.geo.country` to alpha-3 (via `Locale.getISO3Country()`, full ISO coverage) **only when `ALPHA3` is selected**.
- Default is planned to flip to `ALPHA3` in **4.0** (noted in the Javadoc).

## Test plan
- `GeoLocationParameterBuilderTest.testToAlpha3_convertsAlpha2AndPassesThroughAlpha3` covers the conversion helper.
- [ ] Follow-up from review: add a test exercising the flag (ALPHA2 vs ALPHA3 output), and address the inline edge-case notes (null/unmappable → omit, ISO-3 validation, locale-safe upper-casing).

Closes #872


mdanylov-sigma:
Hey @nsp37! Could you check comments in the issue #872 and align your implementation with new requirements ?

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/GeoLocationParameterBuilder.java
```diff
@@ -84,12 +85,48 @@ private void setLocation(AdRequestInput adRequestInput, LocationInfoManager loca
...
                 }
 
+                // OpenRTB device.geo.country is ISO-3166-1 alpha-3, but the
+                // telephony (getSimCountryIso / getNetworkCountryIso) and Geocoder
+                // (Address.getCountryCode) sources return alpha-2 (e.g. "US"); only
+                // the Locale.getISO3Country() fallback was already alpha-3.
+                // Opt-in via PrebidMobile.setGeoCountryFormat(ALPHA3) so we don't
+                // silently change existing behavior; defaults to alpha-2 (planned
+                // to default to alpha-3 in 4.0). Idempotent for values already alpha-3.
+                if (PrebidMobile.getGeoCountryFormat() == GeoCountryFormat.ALPHA3) {
```
mdanylov-sigma, line 95: In the default `ALPHA2` mode this block doesn't run, so whatever the fallbacks produced is sent as-is, and the locale fallback (L78-L79) already returns alpha-3. For example, a Wi-Fi-only tablet with no SIM and an en-US locale sends `"country":"USA"` while the flag says alpha-2.

#872 asked for a flag *telling which format will be used*, so I think `ALPHA2` should normalize to alpha-2 as well (e.g. a `toAlpha2()` counterpart in an `else` branch). If the intent is to keep today's mixed output, `ALPHA2` should be renamed or documented as the legacy, non-normalized behavior instead.

@YuriyVelichkoPI, which one did you have in mind? Normalizing would change what the locale path sends today (`USA` → `US`).

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/GeoLocationParameterBuilder.java
```diff
@@ -84,12 +85,48 @@ private void setLocation(AdRequestInput adRequestInput, LocationInfoManager loca
...
+        String c = country.trim().toUpperCase(Locale.ROOT);
+        if (c.length() == 3) {
+            return c; // already alpha-3
+        }
+        if (c.length() != 2) {
+            return "";
+        }
+        try {
+            String iso3 = new Locale("", c).getISO3Country();
+            return iso3 != null ? iso3 : "";
```
mdanylov-sigma: Returning `""` doesn't omit the field: `Geo.getJsonObject()` uses `putOpt`, which only skips `null`, so the request carries `"country":""`.

In `ALPHA3` mode that's a small regression:
- a `null` from `Address.getCountryCode()` used to omit the field and now sends `""`;
- a code the tables can't map, such as Kosovo (`XK`), used to be sent as `"XK"` and now goes out as `""`.

Could `toAlpha3` return `null` for anything it can't convert (with the test switched to `assertNull`)?

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/GeoLocationParameterBuilder.java
```diff
@@ -84,12 +85,48 @@ private void setLocation(AdRequestInput adRequestInput, LocationInfoManager loca
...
+     * "GB" -> "GBR") via the JDK's own ISO tables — full coverage. Returns a value
+     * already alpha-3 unchanged, and "" for empty / unknown / unconvertible input
+     * (so the caller emits no country rather than a malformed one).
+     */
+    static String toAlpha3(String country) {
+        if (country == null) {
+            return "";
+        }
+        String c = country.trim().toUpperCase(Locale.ROOT);
+        if (c.length() == 3) {
```
mdanylov-sigma, line 136: This accepts any 3-character string as alpha-3. On Android, `Locale.getISO3Country()` returns a 3-character region unchanged. So on a device with an `es-419` (Latin American Spanish) locale and no SIM country, the locale fallback should produce `"419"`, a UN M.49 code, and `ALPHA3` mode passes it straight through.

I haven't reproduced this on a device, and the Robolectric test can't catch it because the JVM's `Locale` throws for `419` instead of returning it.

Suggestion: validate against the real ISO-3 set, e.g. a static `Set<String>` built from `Locale.getISOCountries()` mapped through `new Locale("", cc).getISO3Country()`.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/GeoLocationParameterBuilder.java
```diff
@@ -84,12 +85,48 @@ private void setLocation(AdRequestInput adRequestInput, LocationInfoManager loca
...
+     * Convert an ISO-3166-1 alpha-2 country code to alpha-3 (e.g. "US" -> "USA",
+     * "GB" -> "GBR") via the JDK's own ISO tables — full coverage. Returns a value
+     * already alpha-3 unchanged, and "" for empty / unknown / unconvertible input
+     * (so the caller emits no country rather than a malformed one).
+     */
+    static String toAlpha3(String country) {
+        if (country == null) {
+            return "";
+        }
+        String c = country.trim().toUpperCase(Locale.ROOT);
```
mdanylov-sigma, line 135: `Locale.ROOT` is right here, but the value can already be broken by the time it arrives: `getTelephonyCountry()` (L134-L135) calls `toUpperCase()` without a locale.

On a device whose default locale is Turkish (or Azerbaijani), `"it"` becomes `"İT"` (dotted İ). `ALPHA2` then sends `"İT"` and `ALPHA3` sends `""`. The same happens for IN, ID, IE, IL, IR, IQ, IS, FI and others.

Could you switch those two calls to `toUpperCase(Locale.ROOT)` while you're in this file?

PrebidMobile/PrebidMobile-core/src/test/java/org/prebid/mobile/rendering/networking/parameters/GeoLocationParameterBuilderTest.java
```diff
@@ -134,4 +134,20 @@ public void testLocationPrecisionNoLimit() throws Exception {
         assertEquals(LATITUDE.floatValue(), adRequestInput.getBidRequest().getDevice().getGeo().lat, 0.0f);
         assertEquals(LONGITUDE.floatValue(), adRequestInput.getBidRequest().getDevice().getGeo().lon, 0.0f);
     }
+
+    @Test
+    public void testToAlpha3_convertsAlpha2AndPassesThroughAlpha3() {
```
mdanylov-sigma, line 141: This covers the helper, but nothing tests the flag itself: removing or inverting the `if` on L95 of the builder keeps the whole suite green.

The existing tests in this class never reach the country code either. They don't set `PrebidContextHolder` or a SIM country, so `country` stays `null` in all of them (the expected JSON has no `country`). Could you add:
- a builder test that sets `PrebidContextHolder.setContext(...)` and `shadowOf(telephonyManager).setSimCountryIso("us")`, then asserts `"US"` by default and `"USA"` with `ALPHA3`;
- an `@After` that resets `PrebidMobile.setGeoCountryFormat(GeoCountryFormat.ALPHA2)`, since the static state otherwise leaks into other tests;
- a default/`null` case for the new setter in `PrebidMobileTest`.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/GeoCountryFormat.java
```diff
@@ -0,0 +1,38 @@
...
+ *    limitations under the License.
+ */
+
+package org.prebid.mobile;
+
+/**
+ * Format used for the auto-detected {@code device.geo.country} in the bid
+ * request. See {@link PrebidMobile#setGeoCountryFormat(GeoCountryFormat)}.
+ */
+public enum GeoCountryFormat {
```
mdanylov-sigma, line 23: A timing question before this becomes public API. In #872 Yuriy also asked whether the SDK should have a toggle to not send country at all, and that's still unanswered.

If the answer is yes, it would land either as a third value here (e.g. `NONE`) or as a separate flag next to `setGeoCountryFormat`. Reshaping or renaming a public enum after release is a breaking change.

@YuriyVelichkoPI @justadreamer, is it worth settling this before merge?

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/GeoLocationParameterBuilder.java
```diff
@@ -84,12 +85,48 @@ private void setLocation(AdRequestInput adRequestInput, LocationInfoManager loca
                     geo.country = list.get(0).getCountryCode();
```
mdanylov-sigma, line 85: Not blocking, and this is existing code: `Geocoder.getFromLocation` (L83-L84) is a synchronous network call. This builder runs from `Requester.makeAdRequest()` → `URLBuilder.buildUrl()` on the caller's thread, which is usually the main thread via `fetchDemand`.

It's rarely reached today because the locale fallback almost always fills the value first. I'm only flagging it in case reverse geocoding moves earlier (the source the #872 reporter preferred). It would then need the async `GeocodeListener` API (API 33+) or to run off the main thread.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidMobile.java
```diff
@@ -247,6 +249,26 @@ public static boolean isShareGeoLocation() {
...
+    /**
+     * Format for the auto-detected {@code device.geo.country} sent in the bid
+     * request. The OS telephony/geocoder sources return ISO-3166-1 alpha-2
+     * ("US") while the OpenRTB spec requires alpha-3 ("USA").
+     *
+     * Defaults to {@link GeoCountryFormat#ALPHA2} to preserve existing behavior;
+     * set {@link GeoCountryFormat#ALPHA3} to send the spec-compliant format. The
+     * default is planned to change to {@code ALPHA3} in Prebid SDK 4.0.
+     */
+    public static void setGeoCountryFormat(GeoCountryFormat format) {
```
mdanylov-sigma: nit: could you add `@Nullable` to the parameter and `@NonNull` to the getter, like the neighbouring API (`setCustomHeaders(@Nullable ...)`)? Also worth a line in the Javadoc that `null` resets the value to `ALPHA2`.

mdanylov-sigma (commented):
Thanks for adding the config flag so quickly, @nsp37! I checked the update against the requirements @YuriyVelichkoPI set in #872 ([comment](https://github.com/prebid/prebid-mobile-android/issues/872#issuecomment-5752350044)). The overall shape matches: the conversion is opt-in, the default is alpha-2, and the Javadoc mentions the 4.0 flip. The inline comments cover a few gaps. The main ones:

- In the default `ALPHA2` mode the SDK can still send alpha-3 (from the locale fallback), so the flag doesn't control the format yet.
- In `ALPHA3` mode a few inputs still produce values that aren't alpha-3 (`""`, `"419"`, codes from Turkish-locale devices).
- The tests don't exercise the flag itself.

Two things outside the diff:

1. **PR description:** it still describes an unconditional alpha-3 conversion. Could you update it to cover the `GeoCountryFormat` opt-in, the alpha-2 default and the 4.0 plan, and add `Closes #872`?
2. **How reliable the country signal is** (existing code, so not blocking this PR, but it bears on Yuriy's question in #872 about whether carrier data is accurate):
   - `getTelephonyCountry()` prefers `getSimCountryIso()` over `getNetworkCountryIso()` ([L137-L141](https://github.com/prebid/prebid-mobile-android/blob/d39cc0d4f7e384a1b1713763cb2c399999bc4f17/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/GeoLocationParameterBuilder.java#L137-L141)). For a roaming user that gives the SIM's home country, not the country the device is in, so it can contradict `lat`/`lon`. Checking the network country first would fix that.
   - The locale fallback ([L77-L80](https://github.com/prebid/prebid-mobile-android/blob/d39cc0d4f7e384a1b1713763cb2c399999bc4f17/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/GeoLocationParameterBuilder.java#L77-L80)) reports the region of the UI language, not the device's location. For example, an en-US tablet in Germany sends `USA`.

   Probably a follow-up rather than part of this PR, but worth agreeing on.

nsp37:
I'll get the hang of these eventually — thanks for the guidance so far!

On the reliability question: we don't have access to this on iOS, but I lean toward trusting the carrier/network data here. Now that alpha-3 is in place, being able to derive the country for publishers is a meaningful feature and a nice end-engineer simplification — as long as it isn't violating Prebid's desired rules. In our own implementation/fork we also check a set of remote CDN headers to try to fill this; I left that out of these PRs.

Happy to align on the open items (e.g. whether `ALPHA2` should normalize to alpha-2 vs. remain legacy passthrough, and the network-country-first / validation cleanups) once you and @YuriyVelichkoPI settle on the intended semantics.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/GeoLocationParameterBuilder.java
nsp37: Fixed in b0dfc0b — `toAlpha3` now returns `null` (not `""`) for empty/unknown/unconvertible input, so `putOpt` omits the field. Covers a null `Address.getCountryCode()` and unmappable codes like Kosovo `XK`.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/GeoLocationParameterBuilder.java
nsp37: Fixed in b0dfc0b — 3-char inputs are now validated against the real ISO-3 set (built once from `Locale.getISOCountries()` mapped via `getISO3Country()`), so a UN M.49 code like `419` is rejected (→ `null`/omitted) instead of passed through as "alpha-3".

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/networking/parameters/GeoLocationParameterBuilder.java
nsp37: Fixed in b0dfc0b — `getTelephonyCountry()` now upper-cases with `Locale.ROOT`, so Turkish/Azerbaijani-locale devices no longer corrupt `IT/IN/ID/IE/IL/...`.

PrebidMobile/PrebidMobile-core/src/test/java/org/prebid/mobile/rendering/networking/parameters/GeoLocationParameterBuilderTest.java
nsp37: Added a flag set/get test (incl. `null` → default `ALPHA2`) in b0dfc0b. A test that the *builder* honors the flag needs telephony/geocoder mocking to force a country source — happy to add it, likely alongside the `ALPHA2`-normalize decision above since that changes what the ALPHA2 branch should assert.

PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/PrebidMobile.java
nsp37: Done in 429aebe — `@Nullable` on the setter param and `@NonNull` on the getter, matching the neighbouring config API.

PrebidMobile/PrebidMobile-core/src/test/java/org/prebid/mobile/rendering/networking/parameters/GeoLocationParameterBuilderTest.java
nsp37: Added builder-level tests in 68b126d that set `PrebidContextHolder` and a shadowed SIM/network country of `us`: the default format sends `"US"` and `ALPHA3` sends `"USA"`, so inverting or removing the gate now fails. The empty-country case (e.g. an unmappable locale throwing in `getISO3Country()`) now omits the field instead of sending `"country":""`.

